PR #254, opened 2014-11-24
Changed mon restart during an upgrade sequence to fix #10178

yuriw:
For suites firefly-x and giant-x parallel only

@liewegas pls review


dachary:
@yuriw is this still necessary ?

PR #283, opened 2015-01-08
Use qemu-kvm in the qemu task for rpm based systems.

andrewschoen:
This resolves these two tracker issues:

http://tracker.ceph.com/issues/10275
http://tracker.ceph.com/issues/10276

Even though we are now using qemu-kvm, the version that is install wasn't compiled with rbd support so qemu tests will still fail when starting up qemu.  I've cut another ticket (http://tracker.ceph.com/issues/10480) to fix this as it can't be done in ceph-qa-suite.

### DIFF
--- a/tasks/qemu.py
+++ b/tasks/qemu.py
@@ -169,14 +169,20 @@ def run_qemu(ctx, config):
                 ]
             )
 
-        base_file = '{tdir}/qemu/base.{client}.qcow2'.format(tdir=testdir, client=client)
+        base_file = '{tdir}/qemu/base.{client}.qcow2'.format(
+            tdir=testdir,
+            client=client
+        )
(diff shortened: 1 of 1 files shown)


tasks/qemu.py
```diff
@@ -169,14 +169,21 @@ def run_qemu(ctx, config):
                 ]
             )
 
-        base_file = '{tdir}/qemu/base.{client}.qcow2'.format(tdir=testdir, client=client)
+        base_file = '{tdir}/qemu/base.{client}.qcow2'.format(
+            tdir=testdir,
+            client=client
+        )
+        system_type = teuthology.get_system_type(remote)
```
zmc: remote.os.package_type!

PR #284, opened 2015-01-09
Fix mpi calls

wusui:
added missing -o option to mpi exec.
added gmnt file reference to mpi exec.
run fsx-mpi as sudo.
run chef task so that libmpich2-dev gets installed (fsx-mpi.c uses it).

Signed-off-by: Warren Usui warren.usui@inktank.com
Fixes: #1398

### DIFF
--- a/suites/fs/multiclient/tasks/fsx-mpi.yaml.disabled
+++ b/suites/fs/multiclient/tasks/fsx-mpi.yaml.disabled
@@ -3,6 +3,7 @@ os_type: ubuntu
 os_version: "14.04"
 
 tasks:
+- chef: null
 - pexec:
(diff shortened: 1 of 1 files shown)


zmc:
ok to test

dachary:
@wusui ping ?

liewegas:
need to drop the chef: line and rename the yaml so it's not .disabled

liewegas:
@batrick probably just want to apply the oneliner portion of this to master?

batrick:
Sure. Keep it disabled?

liewegas:
if it works we should enable it.. but it was disabled bc it didn't work.. 
:/

PR #296, opened 2015-01-14
rados.py: Add write_fadvise_dontneed field

majianpeng:
Signed-off-by: Jianpeng Ma <jianpeng.ma@intel.com>


liewegas:
can you also patch a few files in suites/rados/thrash/workloads so that they set the write_fadvise_dontneed option?

majianpeng:
OK, i'll do that.

majianpeng:
Add a file to do write with write_fadvise_dontneed=ture. plz review. Thanks!

PR #302, opened 2015-01-19
ceph: set crush tunables

liewegas:
Like so:

overrides:
  ceph:
    crush_tunables: hammer

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -145,6 +145,16 @@ def valgrind_post(ctx, config):
         if valgrind_exception is not None:
             raise valgrind_exception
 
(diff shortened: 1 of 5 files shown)


liewegas:
tested!

tasks/ceph.py
```diff
@@ -145,6 +145,16 @@ def valgrind_post(ctx, config):
         if valgrind_exception is not None:
             raise valgrind_exception
 
+@contextlib.contextmanager
+def other_setup(ctx, config):
```
zmc: Why call it "crush_tunables" ? Whatever the name, it needs some docs!

PR #303, opened 2015-01-19
tasks/mds_flush: use new stray/purge perfcounters

jcsp:
[Depends on Ceph PR https://github.com/ceph/ceph/pull/3402]

Where previously we only tracked RADOS-level delete
ops during deletion, now also verify that they
correspond to the right number of MDS-level purge
operations.


jcsp:
The perfcounters are in master, so this should be good to go (have tested running it by hand)

jcsp:
aaand I just closed and reopened this because apparently I can't use a mouse

PR #305, opened 2015-01-21
Fixes http://tracker.ceph.com/issues/10593

yuriw:
@yehudasa 
@dachary 

Pls review and merge.

Signed-off-by: Yuri Weinstein <yuriw@magna002.ceph.redhat.com>

### DIFF
--- a/suites/upgrade/dumpling/rgw/2-workload/testrgw.yaml
+++ b/suites/upgrade/dumpling/rgw/2-workload/testrgw.yaml
@@ -2,5 +2,6 @@ workload:
   sequential:
   - s3tests:
       client.0:
+         force-branch: dumpling
          rgw_server: client.0
 


yehudasa:
@yuriw I don't think this is the way to go. While it might solve the specific issue you have here, the question is why it doesn't take the suite_branch. Otherwise whenever we branch out we need to remember to change this, and that's definitely not a good solution.

yuriw:
@yehudasa we know why it's not defaulting to suite_branch and http://tracker.ceph.com/issues/10593 and https://github.com/ceph/ceph-qa-suite/blob/master/tasks/s3tests.py#L125
I will keep issue http://tracker.ceph.com/issues/10593 open and it will have to be fixed in teuthology code

yuriw:
@yehudasa FYI https://github.com/ceph/ceph-qa-suite/pull/306

PR #310, opened 2015-01-21
drop tiobench tests

gregsfortytwo:
The tiobench software has been abandoned upstream for years. Fedora and
Debian are no longer shipping the tiobench package, so we've had to
carry the package ourselves in the Ceph project, and we're trying to
slim down our dependencies where it makes sense to do so.

Nuke the tiobench tests.

http://tracker.ceph.com/issues/10152 Refs: #10152

Signed-off-by: Ken Dreyer <kdreyer@redhat.com>
(cherry picked from commit 8a317125fc542667e57c66017db6e80a3edc24f2)

Conflicts:
	suites/fs/basic/tasks/cfuse_workunit_suites_tiobench.yaml
	suites/fs/samba/workload/cifs-tiobench.yaml
	suites/kcephfs/cephfs/tasks/kclient_workunit_suites_tiobench.yaml

Note in particular that this drops suites/upgrade-fs/fs/5-next-workload/tiobench.yaml. I think that should be okay (the 5 is just for ordering and it's fine to skip, right?) but I don't really know how the upgrade tests function.

### DIFF
--- a/suites/fs/basic/tasks/cfuse_workunit_suites_tiobench.yaml
+++ b/suites/fs/basic/tasks/cfuse_workunit_suites_tiobench.yaml
@@ -1,8 +1,0 @@
-tasks:
-- install:
-- ceph:
-- ceph-fuse:
-- workunit:
-    clients:
-      all:
-        - suites/tiobench.sh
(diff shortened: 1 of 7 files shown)


gregsfortytwo:
This was put in manually.

PR #313, opened 2015-01-23
teuthology: some suites still use http://ceph.newdream.net

armaseg:
This probably redirects to http://ceph.com but ceph.newdream.net still appears in some places

http://tracker.ceph.com/issues/9922 Fixes: #9922

Signed-off-by: Armando Segnini  <armando.segnini@telecom-bretagne.eu>


zmc:
Thanks for the patch!

The commit msg isn't quite accurate though; any chance you could change it to something like 'use ceph.com git mirror' ?

PR #314, opened 2015-01-23
ceph: test pg removal interruption; add ceph daemon failur einjection support

liewegas:


liewegas:
needs this in teuthology: https://github.com/ceph/teuthology/pull/419

PR #316, opened 2015-01-27
mds_scrub_checks: only direct ops at the active MDS

gregsfortytwo:
Change the config option from mds_id to mds_rank to reflect the
fact that it's the rank we want to make use of (and will continue
to want when we're doing stuff like force exporting from one rank
to another).

Fixes: #10361


jcsp:
Merged -- it does make me think that in the future we might find in the future that we want a more general "asok-by-rank" function that doesn't make the assumption that all ranks are up:active.

PR #317, opened 2015-01-28
rbd: load tiobench from the dumpling branch

dachary:
It's not generic but tiobench tests have been removed after giant.


dachary:
@dillaman or @jdurgin does it look like a sensible way to address http://pulpito.ceph.com/loic-2015-01-27_17:07:21-rbd-dumpling-backports---basic-multi/725453/ ? This is in the context of http://tracker.ceph.com/issues/10560

jdurgin:
We should probably cut it from the dumpling test suite too - already done on firefly and giant in d226f34d9c62cbaf6f72681b00f32406a3d12edf . I don't think it's ever uncovered any rbd bugs that the other qemu tests didn't catch first, and as Ken's commit notes, it's no longer maintained.

dachary:
@jdurgin thanks for the quick review, I picked @ktdreyer patch at https://github.com/ceph/ceph-qa-suite/pull/318 instead

PR #323, opened 2015-02-04
Fixes #10500 for v0.80.4 and higher

yuriw:
@athanatos Pls review


athanatos:
I'm a bit confused about this.  Does this mean that the previous test had

osd map max advance

set to 1000?

If that's the case, I'm wrong about what caused the bug.

yuriw:
OK let's wait then for more evidence .

PR #332, opened 2015-02-12
Fixes #10869, added {role} to the dir name that stores the workunits.list

yuriw:
@jdurgin Pls take a look and then I will send it to Zack

### DIFF
--- a/tasks/workunit.py
+++ b/tasks/workunit.py
@@ -308,13 +308,13 @@ def _run_tests(ctx, refspec, role, tests, env, subdir=None, timeout=None):
             'if', 'test', '-e', 'Makefile', run.Raw(';'), 'then', 'make', run.Raw(';'), 'fi',
             run.Raw('&&'),
             'find', '-executable', '-type', 'f', '-printf', r'%P\0'.format(srcdir=srcdir),
-            run.Raw('>{tdir}/workunits.list'.format(tdir=testdir)),
+            run.Raw('>{tdir}/workunits.list.{role}'.format(tdir=testdir, role=role)),
         ],
     )
 
     workunits = sorted(misc.get_file(
         remote,
-        '{tdir}/workunits.list'.format(tdir=testdir)).split('\0'))
+        '{tdir}/workunits.list.{role}'.format(tdir=testdir, role=role)).split('\0'))
     assert workunits
 
     try:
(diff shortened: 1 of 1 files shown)


yuriw:
@zmc   Pls review, it fixes http://tracker.ceph.com/issues/10869

tasks/workunit.py
```diff
@@ -308,13 +308,13 @@ def _run_tests(ctx, refspec, role, tests, env, subdir=None, timeout=None):
             'if', 'test', '-e', 'Makefile', run.Raw(';'), 'then', 'make', run.Raw(';'), 'fi',
             run.Raw('&&'),
             'find', '-executable', '-type', 'f', '-printf', r'%P\0'.format(srcdir=srcdir),
-            run.Raw('>{tdir}/workunits.list'.format(tdir=testdir)),
+            run.Raw('>{tdir}/workunits.list.{role}'.format(tdir=testdir,role=role)),
```
zmc: Needs a space between args, e.g. ``format(tdir=testdir, role=role)``

tasks/workunit.py
```diff
@@ -308,13 +308,13 @@ def _run_tests(ctx, refspec, role, tests, env, subdir=None, timeout=None):
...
         ],
     )
 
     workunits = sorted(misc.get_file(
         remote,
-        '{tdir}/workunits.list'.format(tdir=testdir)).split('\0'))
+        '{tdir}/workunits.list.{role}'.format(tdir=testdir,role=role)).split('\0'))
```
zmc: Needs a space between args, e.g. ``format(tdir=testdir, role=role)``

tasks/workunit.py
```diff
@@ -368,6 +368,6 @@ def _run_tests(ctx, refspec, role, tests, env, subdir=None, timeout=None):
         remote.run(
             logger=log.getChild(role),
             args=[
-                'rm', '-rf', '--', '{tdir}/workunits.list'.format(tdir=testdir), srcdir,
+                'rm', '-rf', '--', '{tdir}/workunits.list.{role}'.format(tdir=testdir,role=role), srcdir,
```
zmc: Needs a space between args, e.g. ``format(tdir=testdir, role=role)``

zmc:
Mainly style things here. I made a few identical comments that should be addressed. Also, the commit message is too long to fit in most interfaces - 50 chars max, and in a paragraph below that you can put all the detail you need.

Finally, are you sure you just want to put this into giant? It will never automatically get into master if that's the case.

yuriw:
@zmc - I hope fixed all spaces.  Re: master, I was planning to do cherry-pick after this one is merged.  Do we use a different way?  If we do, I'd be happy to follow it.

zmc:
@yuriw Thanks for making those changes.

Normally we commit to master and then cherry-pick from there; in the future let's stick to that but this time I think we can make an exception since I don't think it will actually hurt anything.

yuriw:
@zmc cherry-picked to master 9942440ddf9f6847e76de3b49f1ad0d98c62a3d0

PR #333, opened 2015-02-13
Mirror the supported distros used in nightlies in the teuthology suite

andrewschoen:
I decided not to symlink because I'm using a file name convention to
easily filter out distros that need a vps.


andrewschoen:
@yuriw I was already testing against Ubuntu 14.04 and rhel 7.0, these are just the additional distros needed to match up with the nightlies.

PR #337, opened 2015-02-21
lsof on umount failure; catpure logs for aggressively

liewegas:
this is trying to track down why we keep seeing that umount hting all the
time!

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -761,16 +761,24 @@ def cluster(ctx, config):
         for remote, dirs in devs_to_clean.iteritems():
             for dir_ in dirs:
                 log.info('Unmounting %s on %s' % (dir_, remote))
-                remote.run(
-                    args=[
-                        'sync',
-                        run.Raw('&&'),
-                        'sudo',
-                        'umount',
-                        '-f',
-                        dir_
+                try:
+                    remote.run(
+                        args=[
+                            'sync',
+                            run.Raw('&&'),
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
You need to set up PATH or something for the lsof or it will fail on RHEL. :)

liewegas:
fixed!  also added sudo and tested, looks good

tasks/ceph.py
```diff
@@ -76,6 +76,40 @@ def ceph_log(ctx, config):
...
+
+            log.info('Archiving logs...')
+            path = os.path.join(ctx.archive, 'remote')
+            os.makedirs(path)
+            for remote in ctx.cluster.remotes.iterkeys():
+                sub = os.path.join(path, remote.shortname)
+                os.makedirs(sub)
+                teuthology.pull_directory(remote, '/var/log/ceph',
+                                          os.path.join(sub, 'log'))
         pass
```
andrewschoen: Don't need this anymore, right?

tasks/ceph.py
```diff
@@ -728,16 +762,24 @@ def first_in_ceph_log(pattern, excludes):
...
+                        args=[
+                            'sync',
+                            run.Raw('&&'),
+                            'sudo',
+                            'umount',
+                            '-f',
+                            dir_
                         ]
                     )
+                except:
```
andrewschoen: nitpick: bare except

PR #341, opened 2015-02-24
workunit: include /usr/sbin in the PATH for all commands

gregsfortytwo:
Fixes: #10600

Backport: hammer, giant, firefly, dumpling


gregsfortytwo:
This is incredibly simple on its own, but other people use the workunits so I don't want to rush through pushing it into the whole tree.

PR #343, opened 2015-02-24
Use an NFS mount instead of virtio-9p in the qemu task

andrewschoen:
RHEL 7.0 doesn't have virtio-9p, so we switched to using an nfs mount so that the guest can share files with the host during the qemu tasks.  This will work across all distros.

Tested in octo for both rhel and ubuntu.

http://pulpito.ceph.redhat.com/qemu-rhel-issue-10680/
http://pulpito.ceph.redhat.com/qemu-ubuntu-issue-10680/

### DIFF
--- a/tasks/qemu.py
+++ b/tasks/qemu.py
@@ -53,12 +53,18 @@ def generate_iso(ctx, config):
     testdir = teuthology.get_testdir(ctx)
     for client, client_config in config.iteritems():
         assert 'test' in client_config, 'You must specify a test to run'
+        (remote,) = ctx.cluster.only(client).remotes.keys()
         src_dir = os.path.dirname(__file__)
         userdata_path = os.path.join(testdir, 'qemu', 'userdata.' + client)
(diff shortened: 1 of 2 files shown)


tasks/qemu.py
```diff
@@ -154,6 +159,78 @@ def download_image(ctx, config):
                     ],
                 )
 
+
+def _setup_nfs_mount(remote, client, mount_dir):
+    """
+    Sets up an nfs mount on the remote so that the guest
+    can move it's logs over to the remote after the tests
```
zmc: I feel silly saying this, but: "its".

zmc: Also, the guest should be writing logs directly to the nfs mount, not moving them over afterward. That way if the guest blows up, we still have logs.

andrewschoen: Will fix.

andrewschoen: As discussed in standup, it does actually write some logs directly to the nfs share and some it copies over at the end.  I'll make these docs more clear.

PR #355, opened 2015-02-28
all tests must end with .t

dachary:
The hb=firelfy cannot end the URL because of this constraint.


dachary:
test passed http://pulpito.ceph.com/loic-2015-02-28_19:23:35-rbd-firefly-backports---basic-multi/

PR #356, opened 2015-03-02
tasks/cephfs: don't run iptables in parallel

jcsp:
Where multiple MDSs were on the same node, trying
to concurrently update their firewall state was
causing an exception because the iptables command
errors out if another instance is already running.

Fixes: #10948

### DIFF
--- a/tasks/cephfs/filesystem.py
+++ b/tasks/cephfs/filesystem.py
@@ -192,17 +192,27 @@ class Filesystem(object):
         else:
             return self.mds_ids[0]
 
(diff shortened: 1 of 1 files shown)


jcsp:
ok to test

gregsfortytwo:
This looks good, just waiting for tests to pass.

gregsfortytwo:
ok to test

jcsp:
test this please

alfredodeza:
test this please

PR #357, opened 2015-03-02
tasks/cephfs: clean up core on deliberate crash

jcsp:
To avoid internal.coredump task synthesizing a failure
during teardown from the core we left behind.

Fixes: #10949

### DIFF
--- a/tasks/mds_journal_repair.py
+++ b/tasks/mds_journal_repair.py
@@ -6,11 +6,15 @@ Test our tools for recovering the content of damaged journals
 import contextlib
 import json
 import logging
+import os
 from textwrap import dedent
 import time
+from StringIO import StringIO
+import re
 from teuthology.orchestra.run import CommandFailedError
 from tasks.cephfs.filesystem import Filesystem, ObjectNotFound, ROOT_INO
 from tasks.cephfs.cephfs_test_case import CephFSTestCase, run_tests
+from teuthology.orchestra import run
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


jcsp:
ok to test

gregsfortytwo:
ok to test

tasks/mds_journal_repair.py
```diff
@@ -213,16 +216,29 @@ def subtrees_assigned():
...
+
+            # Go remove the coredump from the crash, otherwise teuthology.internal.coredump will
+            # catch it later and treat it as a failure.
+            p = self.fs.mds_daemons[crasher_id].remote.run(args=[
+                "sudo", "sysctl", "-n", "kernel.core_pattern"], stdout=StringIO())
+            core_dir = os.path.dirname(p.stdout.getvalue().strip())
+            if core_dir:
+                log.info("Clearing core dir: {0}".format(core_dir))
+                self.fs.mds_daemons[crasher_id].remote.run(args=[
+                    "sudo", "rm", "-f", run.Raw("{0}/*".format(core_dir))
```
gregsfortytwo: I don't think we can blindly remove cores here; it might take other daemon crash information away. If we've got the core pattern and the process ID we ought to be able to identify the correct one to remove, or we can skip the delete if there are multiple files in the directory.

jcsp: Fair point, I've made it more precise now

jcsp:
test this please

gregsfortytwo:
You tested that the straight pid grabbing worked out correctly, right John? I don't know for sure that the messenger nonce will match up in this case (it might).

And now let's see if we can get Jenkins happy...

gregsfortytwo:
ok to test

gregsfortytwo:
test this please

jcsp:
> You tested that the straight pid grabbing worked out correctly, right John? I don't know for sure that the messenger nonce will match up in this case (it might).

Yep, I ran it, it was happy.

gregsfortytwo:
Cool. It all looks good to me. (Pending test runner confusion...*sigh*)

alfredodeza:
retest this please

PR #364, opened 2015-03-06
Implemented "exclude_packages" feature #11043

yuriw:
@jdurgin Pls review

I did only for firefly-x test as dumpling-x was passing, but maybe we should do for it as well?
Here is the latest run - http://pulpito.front.sepia.ceph.com/teuthology-2015-03-04_23:20:02-upgrade:client-upgrade-master-distro-basic-multi/

### DIFF
--- a/suites/upgrade/client-upgrade/firefly-x/basic/2-workload/rbd_cli_import_export.yaml
+++ b/suites/upgrade/client-upgrade/firefly-x/basic/2-workload/rbd_cli_import_export.yaml
@@ -2,6 +2,7 @@ tasks:
 - workunit:
     branch: firefly
     clients:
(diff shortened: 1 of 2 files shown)


suites/upgrade/client-upgrade/firefly-x/basic/2-workload/rbd_cli_import_export.yaml
```diff
@@ -2,6 +2,7 @@ tasks:
 - workunit:
     branch: firefly
     clients:
+      exclude_packages: ['ceph-test', 'ceph-test-dbg']
```
jdurgin: this should be in the install.upgrade configuration, not the workunit configuration

yuriw:
@jdurgin fixed, my is bad, sorry

PR #368, opened 2015-03-16
whitelist shard missing errors

dachary:
If a shard is missing during a radosbench it can safely be
ignored. It will either be recovered and radosbench will complete
successfully. Or there are too many shard missing to recover the object
and radosbench will fail. Either way we can rely on radosbench to
accurately report success or failure.


dachary:
@dzafman do you have time to take a look ?

PR #371, opened 2015-03-19
Add support for ISOs to tasks/calamari_setup

dmick:
Started by @GregMeno, finished by me

### DIFF
--- a/tasks/calamari_setup.py
+++ b/tasks/calamari_setup.py
@@ -255,35 +255,79 @@ def calamari_install(config, cal_svr):
     delete_iceball = False
 
     if iceball_loc.startswith('http'):
-        get_iceball_with_http(iceball_loc, ice_version, ice_distro, '/tmp')
+        iceball_file = get_iceball_with_http(
+            iceball_loc, ice_version, ice_distro, '/tmp'
+        )
         delete_iceball = True
     elif iceball_loc == '.':  # TODO this is a bad sentinel
-        create_iceball(ice_tool_dir, git_icetool_loc, ice_version, version, ice_distro)
-        # TODO move the ICEBALL to /tmp
+        iceball_file = create_iceball(
+            ice_tool_dir, git_icetool_loc, ice_version, version, ice_distro
+        )
+        shutil.move(iceball_file, '/tmp')
+        iceball_file = os.path.join('/tmp', iceball_file)
         delete_iceball = True
+    else:
+        prefix = os.path.join(iceball_loc,
+            'ICE-{0}-{1}'.format(ice_version, ice_distro))
+        for name in [prefix + ext for ext in ('.iso', '.tar.gz')]:
+            if os.path.exists(name):
+                iceball_file = name
+                break
+        else:
+            raise RuntimeError(
+                'Can''t find {0} at {1}'.format(prefix, iceball_loc)
+            )
 
-    gz_file = 'ICE-{0}-{1}.tar.gz'.format(ice_version, ice_distro)
-    lgz_file = os.path.join(destdir, gz_file)
-    cal_svr.put_file(lgz_file, os.path.join('/tmp/', gz_file))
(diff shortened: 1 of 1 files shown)


tasks/calamari_setup.py
```diff
@@ -179,90 +218,117 @@ def calamari_install(config, cal_svr):
 
     The steps here are:
         -- Get the iceball, building it if necessary.
-        -- Copy the iceball to the calamari server, and untarring it.
-        -- Running ice-setup.py on the calamari server.
-        -- Running calamari-ctl initialize.
+        -- Copy the iceball to the calamari server, and untar/mount it.
+        -- Run ice-setup.py on the calamari server.
```
trhoden: perhaps drop the `.py` part here, since it's no longer a script?  The actual command now is `ice_setup`.

tasks/calamari_setup.py
```diff
@@ -179,90 +218,117 @@ def calamari_install(config, cal_svr):
...
+    elif icetype == 'iso':
+        mountpoint = '/mnt/'   # XXX create?
+        ret = cal_svr.run(
+            args=['sudo', 'mount', '-r', remote_iceball_file, mountpoint,]
+        )
+
+    # install ice_setup package
+    args = {
+        'deb':'sudo dpkg -i /mnt/ice-setup*deb',
+        'rpm':'sudo rpm -i /mnt/ice_setup*rpm'
```
trhoden: I think our docs say to use `yum localinstall` instead of `rpm -i`.

tasks/calamari_setup.py
```diff
@@ -179,90 +218,117 @@ def calamari_install(config, cal_svr):
...
+    }.get(cal_svr.system_type, None)
+    if not args:
+        raise RuntimeError('{0}: unknown system type'.format(cal_svr))
+    import pdb; pdb.set_trace()
```
trhoden: really want this?  probably left over from a debug session.

tasks/calamari_setup.py
```diff
@@ -179,90 +218,117 @@ def calamari_install(config, cal_svr):
...
     ice_in = StringIO(icesetdata)
-    ice_setup_io = StringIO()
-    ret = cal_svr.run(args=['sudo', 'python', 'ice_setup.py'], stdin=ice_in,
-                      stdout=ice_setup_io)
-    log.debug(ice_setup_io.getvalue())
-    # Run Calamari-ceph connect.
+    ice_out = StringIO()
+    args = 'sudo ice_setup'
+    if icetype == 'iso':
+        args += ' -d /mnt'
```
trhoden: Do you foresee any need to handle RPMs that aren't signed?  By default, ice_setup will enable GPGCHECK in the .repo files.  if you pass `--nogpg`, it will disable. Useful for testing out beta .iso's that don't have signed RPMs yet.  But I'm not sure that's a usecase you need to worry about here.

Also, there is no capability for that in the .deb packaging yet.

tasks/calamari_setup.py
dmick: yep

tasks/calamari_setup.py
dmick: will do

tasks/calamari_setup.py
dmick: indeed

tasks/calamari_setup.py
dmick: Possibly, but I prefer to cross that when we come to it; for now, I want the tests to be validating the keys/signing

GregMeno:
I find myself really wanting absolute path to ISO, tar.
mostly because the ISO naming scheme that came out in red-velvet is different than our old convention.
I would be happy to write it.
maybe later it's easy for me to just rename the file
What do you think?

GregMeno:
looks good and passes with the rhel7-iso

PR #373, opened 2015-03-21
fs: fix up dd testing again

gregsfortytwo:
We need to change into the directory during the same shell session as we actually
run the dd...

Fixes: #10861

Backport: hammer, giant, firefly

### DIFF
--- a/suites/fs/basic/tasks/cfuse_workunit_suites_truncate_delay.yaml
+++ b/suites/fs/basic/tasks/cfuse_workunit_suites_truncate_delay.yaml
@@ -10,7 +10,6 @@ tasks:
 - ceph-fuse:
 - exec:
     client.0:
-      - cd $TESTDIR/mnt.*
-      - dd if=/dev/zero of=./foo count=100
+      - cd $TESTDIR/mnt.* && dd if=/dev/zero of=./foo count=100
       - sleep 2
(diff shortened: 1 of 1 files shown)


suites/fs/basic/tasks/cfuse_workunit_suites_truncate_delay.yaml
```diff
@@ -10,7 +10,6 @@ tasks:
 - ceph-fuse:
 - exec:
     client.0:
-      - cd $TESTDIR/mnt.*
-      - dd if=/dev/zero of=./foo count=100
+      - cd $TESTDIR/mnt.* && dd if=/dev/zero of=./foo count=100
       - sleep 2
       - truncate --size 0 ./foo
```
trhoden: Knowing nothing about how the exec task works...

if these don't all get wrapped up into the same shell session, does that mean you have to cd into $TESTDIR/mnt.* again before the truncate command?

liewegas: I think they're individual exec calls from teuthology.  so, yep, no working directory

:+1:

gregsfortytwo: *sigh*

Yes, thanks @trhoden.

PR #374, opened 2015-03-21
ceph_manager: Check for exit status 11 from ceph-objectstore-tool import

dzafman:


dzafman:
@yuriw I'm not sure what branches this needs to be merged into.  I'm not familiar with what branches of ceph-qa-suite are involved during upgrade testing.

yuriw:
@dzafman I think commits should go to hammer and master, but I will double check

yuriw:
@dzafman I am not sure if you want me ro merge it or have somebody else to review ?

dzafman:
@yuriw It isn't too complicated.  The python is valid code and it looks just like the other ignore case.

PR #376, opened 2015-03-23
ensure summary is looked for the user we need (part 2)

dachary:
Move the get_user_summary(out, user) logic to util.rgw so that it can be
shared between radosgw_admin_rest.py and radosgw_admin.py and modify
them accordingly.

http://tracker.ceph.com/issues/11180 Fixes: #11180

(cherry picked from commit 97e6d808f086834f2350084d5de36be285ad2bde)

Conflicts:
	tasks/radosgw_admin.py

### DIFF
--- a/tasks/radosgw_admin.py
+++ b/tasks/radosgw_admin.py
@@ -21,18 +21,9 @@ import boto.s3.acl
 import util.rgw as rgw_utils
 
 from teuthology import misc as teuthology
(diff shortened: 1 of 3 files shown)


dachary:
tested at http://pulpito.ceph.com/loic-2015-03-23_00:22:33-rgw-giant---basic-multi

PR #379, opened 2015-03-26
fix watch-notify test

liewegas:


tasks/watch_notify_same_primary.py
```diff
@@ -67,7 +69,29 @@ def start_watch(n):
...
+                        "-p", pool,
+                        "listwatchers",
+                        obj(i)],
+                    stdout=StringIO())
+                lines = proc.stdout.getvalue()
+                n = lines.count('watcher=')
+                log.info('i see %d watchers for %s', n, obj(i))
+                if n >= 1:
+                    break
+                time.sleep(1)
```
alfredodeza: the `safe_while` utility allows you to define how many seconds to sleep (and other options). By default it will try for 10 times sleeping 6 seconds in between.

Was this intended to add another second to `time.sleep` ?

andrewschoen:
+1 to what alfredodeza mentioned, you probably don't want / need the time.sleep

Also, flake8 is complaining.  

./tasks/watch_notify_same_primary.py:106:31: F812 list comprehension redefines 'n' from line 90

PR #380, opened 2015-03-26
Make sure that ulimits are adjusted for ceph-objectstore-tool

dzafman:
Fixes: #11204

Signed-off-by: David Zafman <dzafman@redhat.com>


liewegas:
lgtm

the way the process is invoked looks different than elsewhere (where each arg is called out as a separate string in a list), but adding adjust-ulimits is definitely the right thing to do here.

PR #382, opened 2015-03-27
calamari_setup: der.  Use dict.update() correctly

dmick:


andrewschoen:
Bah, I should have caught this the first time.

PR #383, opened 2015-03-27
CephFS test updates, + add tests for strays + sessionmap

jcsp:


tasks/cephfs/filesystem.py
```diff
@@ -140,16 +140,22 @@ def clear_ceph_conf(self, subsys, key):
...
         status = self.mon_manager.get_mds_status_all()
         for mds_id, mds_status in status['info'].items():
             if mds_status['state'] not in ["up:active", "up:standby", "up:standby-replay"]:
                 log.warning("Unhealthy mds state {0}:{1}".format(mds_id, mds_status['state']))
                 return False
+            elif mds_status['state'] == 'up:active':
+                active_count += 1
 
-        return True
+        return active_count >= status['max_mds']
```
gregsfortytwo: Okay, so we can't actually have more MDSes active than max_mds. I guess if the states are changing while iterating we could count one active logical rank across two daemons, but in that case is the cluster really healthy?

jcsp: If you have two MDSs active, and set max_mds to 1, there you go.  This is intended to work even in that case, where max_mds has been modified.

tasks/cephfs/filesystem.py
```diff
@@ -371,20 +402,34 @@ def wait_for_state(self, goal_state, reject=None, timeout=None, mds_id=None):
...
+            if mds_id is not None:
+                # mds_info is None if no daemon with this ID exists in the map
+                mds_info = self.mon_manager.get_mds_status(mds_id)
+                current_state = mds_info['state'] if mds_info else None
+                log.info("Looked up MDS state for {0}: {1}".format(mds_id, current_state))
+            else:
+                # In general, look for a single MDS
+                mds_status = self.mon_manager.get_mds_status_all()
+                states = [m['state'] for m in mds_status['info'].values()]
+                if [s for s in states if s == goal_state] == [goal_state]:
```
gregsfortytwo: Maybe my python parsing is failing me, but this looks like maybe something got copied too many times?

tasks/cephfs/cephfs_test_case.py
```diff
@@ -15,11 +22,55 @@ class CephFSTestCase(unittest.TestCase):
     Handles resetting the cluster under test between tests.
     """
     # Environment references
+    mounts = None
+    fs = None
+    ctx = None
+
+    # FIXME weird explicit naming
```
gregsfortytwo: Going to fix this?

tasks/cephfs/cephfs_test_case.py
```diff
@@ -15,11 +22,55 @@ class CephFSTestCase(unittest.TestCase):
...
+        if self.REQUIRE_KCLIENT_REMOTE:
+            if not isinstance(self.mounts[0], FuseMount) or not isinstance(self.mounts[1], FuseMount):
+                # kclient kill() power cycles nodes, so requires clients to each be on
+                # their own node
+                if self.mounts[0].client_remote.hostname == self.mounts[1].client_remote.hostname:
+                    raise case.SkipTest("kclient clients must be on separate nodes")
+
+        if self.REQUIRE_ONE_CLIENT_REMOTE:
+            if self.mounts[0].client_remote.hostname in self.fs.get_mds_hostnames():
+                raise case.SkipTest("Require first client to on separate server from MDSs")
```
gregsfortytwo: s/to on/to be on/

tasks/cephfs/cephfs_test_case.py
gregsfortytwo: Actually it looks like maybe you did below with the "Create friendly mount_a, mount_b attrs" section?

tasks/cephfs/filesystem.py
```diff
@@ -445,6 +491,55 @@ def read_backtrace(self, ino_no):
...
+
+        if present:
+            log.info("Objects not absent (ino {0}, size {1}): {2}".format(
+                ino, size, present
+            ))
+            return False
+        else:
+            log.info("All objects for ino {0} size {1} are absent".format(ino, size))
+            return True
+
```
gregsfortytwo: These two functions should probably share the listing implementation? Or just swap which direction with a  parameter; not sure which is cleaner.

tasks/cephfs/test_strays.py
```diff
@@ -0,0 +1,465 @@
...
+        create_script = dedent("""
+            import os
+
+            mount_path = "{mount_path}"
+            subdir = "delete_me"
+            size_unit = {size_unit}
+            file_multiplier = {file_multiplier}
+            os.mkdir(os.path.join(mount_path, subdir))
+            for i in xrange(0, file_multiplier):
+                for size in xrange(0, 16*size_unit, size_unit):
```
gregsfortytwo: The use of 16 as a magic number here (and particularly below) is a bit confusing when skimming. Be good if it was changeable, too?

tasks/cephfs/test_strays.py
```diff
@@ -0,0 +1,465 @@
...
+                    num_strays_purging, num_strays,
+                    total_strays_purged, total_strays_created
+                ))
+                time.sleep(1)
+                elapsed += 1
+
+        background_thread.join()
+
+        # Check that we got up to a respectable rate during the purge.  This is totally
+        # racy, but should be safeish unless the cluster is pathalogically slow, or
```
gregsfortytwo: "pathologically", fyi.

tasks/cephfs/test_strays.py
```diff
@@ -0,0 +1,465 @@
...
+        self.assertEqual(mdc_stats['num_strays'], 0)
+        self.assertEqual(mdc_stats['num_strays_purging'], 0)
+        self.assertEqual(mdc_stats['num_strays_delayed'], 0)
+        self.assertEqual(mdc_stats['num_purge_ops'], 0)
+        self.assertEqual(mdc_stats['strays_created'], total_inodes)
+        self.assertEqual(mdc_stats['strays_purged'], total_inodes)
+
+    def get_mdc_stat(self, name, mds_id=None):
+        return self.fs.mds_asok(['perf', 'dump', "mds_cache", name],
+                                mds_id=mds_id)['mds_cache'][name]
```
gregsfortytwo: You've got a lot of places where you just repeatedly call this function sequentially, and a lot of others where you fetch the perf dump into a local map and then check its members. I doubt it matters, but I can't tell if there's a reason other than "hadn't written this shorthand yet".

tasks/cephfs/test_strays.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        That when doing a rename from A to B, and B has hardlinks,
+        then we make a stray for B which is then reintegrated
+        into one of his hardlinks.
+        """
+        # Create file_a, file_b, and a hardlink to file_b
+
+        # mv file_a file_b
+
+        # Proceed as with test_hardlink_reintegration
```
gregsfortytwo: Uh. :)

tasks/cephfs/test_strays.py
```diff
@@ -0,0 +1,465 @@
...
+
+    def test_migration_on_shutdown(self):
+        """
+        That when an MDS rank is shut down, any not-yet-purging strays
+        are migrated to another MDS's stray dir.
+        """
+
+        # XXX interestingly, export dir on an empty dir doesn't work.
+        # perhaps because it has no actual dirfrag at that stage?
```
gregsfortytwo: Was this note supposed to prompt something else? The "XXX" there seems a little much otherwise.

suites/fs/recovery/tasks/strays.yaml
```diff
@@ -0,0 +1,5 @@
+
+tasks:
+    - cephfs_test_runner:
+        modules:
+            - tasks.cephfs.test_strays
```
gregsfortytwo: The indentation on these task descriptions is varying pretty widely — does it all work? :/ (It looks like maybe you switched to 4-space indentation, but I think all the existing yaml fragments are 2-spaced.)

gregsfortytwo:
Mostly looks good, but some comments and questions.

tasks/cephfs/filesystem.py
jcsp: Do you mean this particular line?  Nothing's jumping out at me as duplicated?

tasks/cephfs/cephfs_test_case.py
jcsp: > Going to fix this?

Probably at some point -- but this is a pre-existing thing, it just used to be duplicated through the cases and now it's done in this test case.

The thing I'm remind myself about here is that it's weird to have mount_a and mount_b instead of just using mounts[0] and mounts[1]

tasks/cephfs/filesystem.py
jcsp: Done - moved the listing into a shared fn

tasks/cephfs/cephfs_test_case.py
jcsp: Done

tasks/cephfs/test_strays.py
jcsp: Yeah, it's pretty arbitrary.  I'll pull it out to class scope so that if we want to make it externally configurable later that'll be easy to to

tasks/cephfs/test_strays.py
jcsp: My brain knew that but apparently my fingers didn't

tasks/cephfs/test_strays.py
jcsp: Some places fetching the overall map and checking members is done to avoid a bunch of round trips, some places it's just because I hadn't written the other one yet (good guess!).  Should be harmless either way

tasks/cephfs/test_strays.py
jcsp: At the time it came as a surprise, now I don't think it's an issue as it just reflects the fact that no dirfrag necessarily exists until some dentry is created.  Will rephrase

jcsp:
This should be good to go - the failures in the last run were unrelated to the changes here

gregsfortytwo:
@jcsp, we need to whitelist the "cluster [WRN] force file system read-only" message for the auto-repair tests. (I assume it's showing up now because of the cluster log, like everything else....)

jcsp:
Added the log whitelist for auto-repair test

PR #384, opened 2015-03-30
suites/fs: enable directory fragmentation

jcsp:
Setting this in tests globally as a precursor
to enabling it by default in the shipping code.
Set a low mds bal split size in order to make
us hit the fragmentation code more often in
than we usually would in a single MDS environment.

### DIFF
--- a/suites/fs/basic/dirfrag/frag_enable.yaml
+++ b/suites/fs/basic/dirfrag/frag_enable.yaml
@@ -1,0 +1,10 @@
+
+overrides:
+    ceph:
+        conf:
+        mds:
+            mds bal frag: true
+            mds bal split size: 100
+            mds bal merge size: 5
+            mds bal split bits: 3
+
--- a/suites/fs/multiclient/dirfrag/frag_enable.yaml
+++ b/suites/fs/multiclient/dirfrag/frag_enable.yaml
@@ -1,0 +1,1 @@
+../../basic/dirfrag/frag_enable.yaml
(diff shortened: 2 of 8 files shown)


gregsfortytwo:
1) Did you try and figure out how many of our tests are actually going to fragment? If we need to set this to 10 in order to get more than one test exercising it we need to know that. ;)
2) It looks like you copied this yaml fragment into each directory, but I think we generally use symlinks to common yaml files like this so that any config changes don't have to get edited in each place?

jcsp:
Updated to use symlinks.

I haven't done the experiment of running through and seeing which if any of the workloads trigger fragmentation with the limit set to 100.  I'm kind of assuming that we have at least some tests that create >100 files in a directory, if not we should add some!  But it will be interesting to do a log grep after running this to see which tests exercised it, and maybe decrease the value if necessary.

gregsfortytwo:
This passed a run: http://pulpito.ceph.com/ubuntu-2015-06-09_15:15:42-fs-greg-fs-testing---basic-multi/

But I haven't looked through the logs to see if it's actually doing much fragmentation.

ukernel:
I think we should set mds_bal_split_size to a smaller value. otherwise the mds will merge the dirfrags immediately after splitting. mds_bal_merge_size  should be smaller than (mds_bal_split_size / (2^mds_bal_split_bits))

ukernel:
ping @jcsp. mds_bal_split_size is 100, which is too small for default mds_bal_merge_size (50) and default mds_bal_split_bits (3).

jcsp:
@ukernel thanks for reminder -- updated the sizes now.

PR #385, opened 2015-03-31
ceph: be less weird about passing -f to mkfs

liewegas:


tasks/ceph.py
```diff
@@ -529,7 +529,8 @@ def cluster(ctx, config):
                     if mount_options is None:
                         mount_options = ['noatime','user_subvol_rm_allowed']
                     if mkfs_options is None:
-                        mkfs_options = ['-m', 'single',
+                        mkfs_options = ['f',
```
andrewschoen: Should this be -f?

liewegas: yep, fixed!

PR #389, opened 2015-04-01
Only use mira and burnupi nodes for ceph-deploy/fs

trhoden:
The test suite uses 3 OSDs, and specifies an override to test on
separate block devices for OSDs and journal, so we need nodes
that have at least 6 block devices free

Signed-off-by: Travis Rhoden <trhoden@redhat.com>

### DIFF
--- a/suites/ceph-deploy/fs/tasks/cfuse_workunit_suites_blogbench.yaml
+++ b/suites/ceph-deploy/fs/tasks/cfuse_workunit_suites_blogbench.yaml
@@ -1,3 +1,5 @@
+machine_type: mira,burnupi
+
 overrides:
   ceph-deploy:
     conf:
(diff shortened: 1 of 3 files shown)


zmc:
I'd be surprised if this worked. Normally this would be done by changing the argument passed to ``teuthology-suite``.

andrewschoen:
@zmc @trhoden This will totally work.  I do the same in the teuthology suite.

https://github.com/ceph/ceph-qa-suite/blob/master/suites/teuthology/no-ceph/distros/vps_rhel6.4.yaml

trhoden:
@andrewschoen I've yet to find an example where this was set to anything besides "vps", and I was concerned since vps is the only thing that's present in the top-level machine_types dir.

andrewschoen:
@trhoden yeah, I think in practice baremetal machine-type is defined with the -suite argument and vps gets the exception in the yaml.  I do know that teuthology ensures all baremetal nodes are the same machine-type, so you might run into issues there.  I suspect it might be ignored for baremetal or try to make all baremetal nodes mira,burnupi.

PR #390, opened 2015-04-01
suites/fs: update log whitelist

jcsp:
These ones had been passing because of #11177,
now that that is fixed they require an update.

Fixes: #11300


suites/fs/recovery/tasks/mds-full.yaml
```diff
@@ -5,6 +5,10 @@ overrides:
             osd:
               osd objectstore: memstore
               memstore device bytes: 100000000
+overrides:
+  ceph:
+    log-whitelist:
+      - OSD full dropping all updates
```
gregsfortytwo: We can't duplicate yaml sections like this; you need to merge the log-whitelist into the previous Ceph one. :)

jcsp: Good spot - done

gregsfortytwo:
Commit got cleaned up a bit and is now in master as fa2a86196c3ed39c1726be689e42ed5d4aae197a.

PR #392, opened 2015-04-02
mkfs.btrfs fails with the -f flag on precise, try without it on failure.

andrewschoen:
See: http://tracker.ceph.com/issues/11308

This is untested and I'm not entirely sure if it's the correct thing to do, but based on the commit (182cb630346344a6cb3773012d460fc8dfd14285) that changed the original functionality I think it's worth it to try without -f on older distros which might not support that flag.

@gregsfortytwo could you give this branch a try on the tests you've been running to see if it helps?

Signed-off-by: Andrew Schoen <aschoen@redhat.com>

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -558,7 +558,19 @@ def cluster(ctx, config):
                         stdout=StringIO(),
                         )
 
-                remote.run(args= ['yes', run.Raw('|')] + ['sudo'] + mkfs + [dev])
+                try:
+                    remote.run(
+                        args=['yes', run.Raw('|')] + ['sudo'] + mkfs + [dev]
+                    )
+                except run.CommandFailedError:
+                    # try again without -f which is not supported on
+                    # older versions of btfs-tools.
+                    log.exception("Trying again without -f...")
+                    mkfs.pop(0)
+                    remote.run(
+                        args=['yes', run.Raw('|')] + ['sudo'] + mkfs + [dev]
+                    )
+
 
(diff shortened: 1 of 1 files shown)


tasks/ceph.py
```diff
@@ -558,7 +558,19 @@ def cluster(ctx, config):
                         stdout=StringIO(),
                         )
 
-                remote.run(args= ['yes', run.Raw('|')] + ['sudo'] + mkfs + [dev])
+                try:
+                    remote.run(
+                        args=['yes', run.Raw('|')] + ['sudo'] + mkfs + [dev]
+                    )
+                except run.CommandFailedError:
```
gregsfortytwo: Don't you want to be a little more specific about this? Like only trying again if it's btrfs, and making sure below that we're removing the "-f" and not something else?
(Somebody someday is going to change that arguments list and they might put it at the front instead of the end.)

andrewschoen: @gregsfortytwo Yeah, we discussed this in standup and I think this change needs a bit more thought.  At this point I'd be +1 on reverting the original PR that introduced this problem so your tests will finish.  We can come back later with a more elegant solution.

gregsfortytwo: I'm currently just reverting in the branch I'm using to test qa-suite
changes with anyway. If somebody else wants to push that into master so
that we aren't seeing failures in the nightlies that's probably wise, but
should take participation from somebody who's going to pay more attention
to than just whining that things aren't working. ;)

liewegas:
my original patch that broke this is reverted, clsoing thing

PR #395, opened 2015-04-07
RBD: add YAML variables to override locations for ceph-qa-chef and xfstests

fullerdj:


jdurgin:
Looks good. One thing: it'd be good to add the new options to the docstrings of 'def task' in each task. That's generally where options are documented in teuthology. What you have in your commit message is good for that.

jdurgin:
LGTM. @andrewschoen @zmc want to take a look?

PR #396, opened 2015-04-08
DNM: fix ceph-deploy task with CephFS (#11316)

jcsp:
DNM because waiting for https://github.com/ceph/ceph-qa-suite/pull/383 to avoid merge discomfort

### DIFF
--- a/tasks/ceph_deploy.py
+++ b/tasks/ceph_deploy.py
@@ -7,12 +7,14 @@ import contextlib
 import os
 import time
 import logging
+import traceback
 
 from teuthology import misc as teuthology
 from teuthology import contextutil
 from teuthology.config import config as teuth_config
 from teuthology.task import install as install_fn
 from teuthology.orchestra import run
+from tasks.cephfs.filesystem import Filesystem
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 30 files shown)


gregsfortytwo:
I didn't look at the ceph-deploy test change too closely, but the rest looks good. I'm including it in a testing run so we can make sure it didn't break anything.

tasks/ceph_deploy.py
```diff
@@ -174,12 +195,12 @@ def build_ceph_cluster(ctx, config):
...
         all_nodes = get_all_nodes(ctx, config)
-        mds_nodes = get_nodes_using_roles(ctx, config, 'mds')
+        mds_nodes = get_nodes_using_role(ctx, 'mds')
         mds_nodes = " ".join(mds_nodes)
-        mon_node = get_nodes_using_roles(ctx, config, 'mon')
+        mon_node = get_nodes_using_role(ctx, 'mon')
         mon_nodes = " ".join(mon_node)
         new_mon = './ceph-deploy new'+" "+mon_nodes
-        install_nodes = './ceph-deploy install '+ceph_branch+" "+all_nodes
+        install_nodes = './ceph-deploy install ' + (ceph_branch if ceph_branch else "--dev=master") + " " + all_nodes
```
trhoden: I only know a little bit about this test suite, but I want to confirm this change.  Right now, if no branch was given it would default to whatever the most recent stable release is (giant, hammer, etc.).  I doubt most tests really rely on this, as you should be explicit what version you want.  However, this will change the default behavior to install master.

I see no problem with that, but it seems like an unrelated change.

Additionally, would this behavior be better captured ten lines up where ceph_branch is initialized and set?  Seems clearer if this was done as an "else" to the "if config.get('branch') is not None"...

jcsp: The reason this change is here is because the existing task would give a nasty exception if there wasn't an explicit branch set in the config -- it's a quick bug fix rather than an outright behavioural change (you only hit this path if you would previously have seen an exception).

The idea of defaulting to master is to match the behaviour of the default ceph install task, i.e. that when you run ceph-qa-suite master without any overrides, it should run against ceph master.

trhoden:
for the ceph-deploy side of things, this looks good to me.  :+1: 

Seems like it is still just waiting on #383

PR #403, opened 2015-04-14
task/samba: use SIGTERM to stop samba server

ukernel:
man samba(8) contains sentences:

To shut down a user's smbd process it is recommended that SIGKILL (-9)
NOT be used, except as a last resort, as this may leave the shared
memory area in an inconsistent state. The safe way to terminate an smbd
is to send it a SIGTERM (-15) signal and wait for it to die on its own.

### DIFF
--- a/tasks/samba.py
+++ b/tasks/samba.py
@@ -164,7 +164,7 @@ def task(ctx, config):
         smbd_cmd = [
                 'sudo',
                 'daemon-helper',
-                'kill',
+                'term',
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
I was testing this (http://pulpito.ceph.com/gregf-2015-04-16_15:01:43-samba-master---basic-multi) and ran into http://tracker.ceph.com/issues/10624 on one of the runs, but the rest of them looked good.

PR #404, opened 2015-04-14
DNM: tasks: s/pending_create/pending_update/ in MDS

jcsp:
DNM: don't merge this until https://github.com/ceph/ceph/pull/4282, it's just an update to keep up with that change.

Tests inspecting snapserver dump need updating
to reflect the rename.


jcsp:
@gregsfortytwo you'll want to pull this on any qa suite branch running against https://github.com/ceph/ceph/pull/4282

PR #416, opened 2015-04-21
Backport to next: fix ceph-deploy task with CephFS (#11316)

jcsp:
Created at Yuri's request -- although this is weird, usually if a next branch is in play we would fix something there and then merge to master, rather than backporting from next.


gregsfortytwo:
Yeah, I'm not sure when next got created or what it started out pointing at, plus things get hard when PRs get stuck in testing for a while. :(

PR #420, opened 2015-04-24
DM - re: issue 11415 and 11414

yuriw:
Initial implementation of mixed-client suite


dachary:
@yuriw is this still valid ?

PR #421, opened 2015-04-28
tasks/cephfs: add test_pool_perm

ukernel:
This tests that client checks pool permission before
copy data to its cache.

### DIFF
--- a/suites/fs/recovery/tasks/pool-perm.yaml
+++ b/suites/fs/recovery/tasks/pool-perm.yaml
@@ -1,0 +1,5 @@
+
+tasks:
+  - cephfs_test_runner:
+      modules:
+        - tasks.cephfs.test_pool_perm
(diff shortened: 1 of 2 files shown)


jcsp:
Looks good apart from a couple of linter gripes:
```
./tasks/cephfs/test_pool_perm.py:4:1: F401 'ObjectNotFound' imported but unused
./tasks/cephfs/test_pool_perm.py:4:1: F401 'ROOT_INO' imported but unused
```

It also needs a .yaml in suites/fs/recovery/tasks so that it runs in the nightlies

jcsp:
:+1: I'm happy with this, let's just have a scheduled run in the lab before merging (@gregsfortytwo are you still doing runs you could cherry pick this onto?)

gregsfortytwo:
Luckily we broke through the backlog so I haven't done any lately. Looks like a run of this branch already got scheduled though!

PR #422, opened 2015-04-28
suites: log whitelist for journal repair test

jcsp:
"bad backtrace on dir ino" is emitted because
journaltool's recover-dentries doesn't write
backtraces.


gregsfortytwo:
Odd, I thought I'd gotten all of these covered. :(

PR #423, opened 2015-04-29
Fix ceph_deploy task with multiple mons

jcsp:


jcsp:
Untested!

yuriw:
Here is a test running http://pulpito-rdu.front.sepia.ceph.com/teuthology-2015-04-29_16:36:38-ceph-deploy:fs-hammer-distro-basic-typica/

PR #425, opened 2015-05-01
cephfs/mount: check FUSE version before invoking flock

gregsfortytwo:
flock only works properly on FUSE versions >=2.9, which is newer
than eg Ubuntu Precise. So check the version on our client mounts and
only test flock if it's at least that new.

### DIFF
--- a/tasks/cephfs/test_client_recovery.py
+++ b/tasks/cephfs/test_client_recovery.py
@@ -5,9 +5,12 @@ Teuthology task for exercising CephFS client recovery
 
 import logging
 import time
+import distutils.version as version
+import re
 
 from teuthology.orchestra.run import CommandFailedError, ConnectionLostError
 from tasks.cephfs.cephfs_test_case import CephFSTestCase
+from teuthology.packaging import get_package_version
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


gregsfortytwo:
I've run the ceph-qa-suite client_recovery job against this with both old and new clients and it behaves correctly. Running a mixed test now just to be sure, but it looks good!

tasks/cephfs/test_client_recovery.py
```diff
@@ -307,19 +308,31 @@ def test_network_death(self):
...
     def test_filelock(self):
         """
         Check that file lock doesn't get lost after an MDS restart
         """
-        lock_holder = self.mount_a.lock_background()
+        a_version = get_package_version(self.mount_a.client_remote, "fuse")
+        b_version = get_package_version(self.mount_b.client_remote, "fuse")
+        flock_version = "2.9"
+        flockable = False
+        if (a_version >= flock_version and b_version >= flock_version):
```
jcsp: String comparison of versions will fail if fuse rev their version to 2.10.  distutils.version.StrictVersion is probably suitable here, although it doesn't like anything other than numbers and dots, so you might also need to truncate the version string to the leading [0-9\.]+

tasks/cephfs/test_client_recovery.py
```diff
@@ -307,19 +310,43 @@ def test_network_death(self):
...
+        b_version_str = get_package_version(self.mount_b.client_remote, "fuse")
+        flock_version_str = "2.9"
+
+        version_regex = re.compile(r"[0-9\.]+")
+        a_result = version_regex.match(a_version_str)
+        self.assertTrue(a_result)
+        b_result = version_regex.match(b_version_str)
+        self.assertTrue(b_result)
+        a_version = version.StrictVersion(a_result.group())
+        self.assertTrue(a_version)
```
jcsp: Minor nitpick: I think these assertTrues on the version aren't meaningful; StrictVersion constructor is guaranteed to give you a valid object, else it would raise an exception (and afaict there's no such thing as StrictVersion that evaluates to false)

jcsp:
:+1: other than the redundant assertTrue([ab]_version) lines

tasks/cephfs/test_client_recovery.py
gregsfortytwo: Do you mean I should zap the checks on the regex as well, or just the StrictVersion constructors? Not being overly familiar with these interfaces I was obviously opting for caution. ;)

PR #426, opened 2015-05-01
add new valgrind tests to rbd suite

dillaman:


tasks/rbd_fsx.py
```diff
@@ -52,7 +52,18 @@ def _run_one_client(ctx, config, role):
...
+        '{tdir}/archive/coverage'.format(tdir=testdir)
+    ])
+
+    if config.get('valgrind'):
+        args = teuthology.get_valgrind_args(
+            testdir,
+            'fsx_{id}'.format(id=role),
+            args,
+            config.get('valgrind')
+        )
```
jdurgin: looks like we need --error-exitcode=1 to detect the errors, or am I missing some generic check of the logs?

jdurgin: jason points out that ceph task's valgrind_post() checks all valgrind logs regardless of ceph task configuration

jdurgin: @dillaman test comment

PR #428, opened 2015-05-04
suites/rados: add test for 11429

athanatos:
This patch also adds some convenience facilities for making
some of the ceph_manager methods into tasks usable from a
yaml file.

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -1902,3 +1902,20 @@ class CephManager:
         Return path to osd data with {id} needing to be replaced
         """
         return "/var/lib/ceph/osd/ceph-{id}"
+
+def utility_task(name):
+    def task(ctx, config):
+        if config is None:
+            config = {}
+        args = config.get('args', [])
+        kwargs = config.get('kwargs', {})
+        fn = getattr(ctx.manager, name)
+        fn(*args, **kwargs)
+    return task
+
+revive_osd = utility_task("revive_osd")
+kill_osd = utility_task("kill_osd")
+create_pool = utility_task("create_pool")
+remove_pool = utility_task("remove_pool")
+wait_for_clean = utility_task("wait_for_clean")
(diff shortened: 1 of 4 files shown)


tasks/ceph_manager.py
```diff
@@ -1902,3 +1902,20 @@ def get_filepath(self):
         Return path to osd data with {id} needing to be replaced
         """
         return "/var/lib/ceph/osd/ceph-{id}"
+
+def utility_task(name):
```
andrewschoen: This needs some docs, please.

tasks/radosbench.py
```diff
@@ -59,12 +61,13 @@ def task(ctx, config):
...
-        if config.get('pool'):
-            pool = config.get('pool')
-            if pool is not 'data':
-                ctx.manager.create_pool(pool, erasure_code_profile_name=profile_name)
-        else:
-            pool = ctx.manager.create_pool_with_unique_name(erasure_code_profile_name=profile_name)
+        if config.get('create_pool', True):
+            if config.get('pool'):
+                pool = config.get('pool')
+                if pool is not 'data':
```
andrewschoen: nitpick: I would prefer != here as we're comparing values not identities.  This should work fine though.

tasks/utility.py
```diff
@@ -0,0 +1,9 @@
+import logging
+import time
+
+log = logging.getLogger(__name__)
+
+def sleep(ctx, config):
```
jdurgin: this kind of thing fits well in teuthology.git/teuthology/misc.py

jdurgin: I see it's a task. Makes sense as a generic teuthology one imho

athanatos:
I want to retest with the changes, don't merge just yet

zmc:
This looks good to me. You mentioned you wanted to test it one more time, so feel free to do so if it works!

PR #429, opened 2015-05-05
tasks: fix intermittent failure in TestFlush

ghost:
This was an overly strict success condition: the
flush operation doesn't promise to leave you an empty
journal, it promises that anything in the journal
before the flush will be flushed.

Fixes: #10712
Signed-off-by: John Spray john.spray@redhat.com
(cherry picked from commit d20d9238cd59bca002d682cbcdc444e15e856de1)

### DIFF
--- a/tasks/mds_flush.py
+++ b/tasks/mds_flush.py
@@ -44,14 +44,42 @@ class TestFlush(CephFSTestCase):
 
         # ...and the journal is truncated to just a single subtreemap from the
         # newly created segment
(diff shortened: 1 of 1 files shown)


dachary:
http://tracker.ceph.com/issues/11492#fs using this branch works http://pulpito.ceph.com/loic-2015-05-05_11:54:53-fs-hammer-backports-testing-basic-multi/

jcsp:
:+1:

PR #432, opened 2015-05-06
suites/rados: add test for 11429 (hammer)

dachary:


dachary:
It passed one successfull run at http://pulpito.ceph.com/loic-2015-05-07_09:34:21-rados-hammer-backports---basic-multi/878540/

PR #433, opened 2015-05-06
Initial check in for upgrades/hammer stable upgrade suite

yuriw:
@liewegas 
@tmuthamizhan 
@jdurgin 

FYI

### DIFF
--- a/suites/upgrade/hammer/2-workload/rbd.yaml
+++ b/suites/upgrade/hammer/2-workload/rbd.yaml
@@ -1,0 +1,14 @@
+workload:
+   sequential:
+   - workunit:
+       clients:
+         client.0:
+           - rbd/import_export.sh
+       env:
+         RBD_CREATE_ARGS: --new-format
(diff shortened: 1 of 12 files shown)


suites/upgrade/hammer/2-workload/rbd.yaml
```diff
@@ -0,0 +1,14 @@
+workload:
+   sequential:
+   - workunit:
+       clients:
+         client.0:
+           - rbd/import_export.sh
+       env:
+         RBD_CREATE_ARGS: --new-format
+   - print: "**** done rbd/import_export.sh 2-workload"
+   - workunit:
```
jdurgin: can run these in parallel. test_cls_rbd won't work if it's upgraded before it starts, since hammer has new class methods it'll try to test

yuriw: are you saying "can't" run these in parallel?

PR #434, opened 2015-05-07
suites/rados: add test for 11429(firefly)

xinxinsh:


dachary:
https://github.com/xinxinsh/ceph-qa-suite/commit/e20aa5bd42c0b8f48605da6cb99f8952109c0244 changes the prototype and will fail, the code does not need to change, it just needs to run conditionally https://github.com/ceph/ceph-qa-suite/pull/435

PR #437, opened 2015-05-08
ensure summary is looked for the user we need (part 2)

dachary:
Move the get_user_summary(out, user) logic to util.rgw so that it can be
shared between radosgw_admin_rest.py and radosgw_admin.py and modify
them accordingly.

http://tracker.ceph.com/issues/11180 Fixes: #11180

(cherry picked from commit 97e6d808f086834f2350084d5de36be285ad2bde)

Conflicts:
	tasks/radosgw_admin.py
        the context around def successful_ops(out): changed but is
	unrelated to the modification

### DIFF
--- a/tasks/radosgw_admin.py
+++ b/tasks/radosgw_admin.py
@@ -21,19 +21,9 @@ import boto.s3.acl
 import util.rgw as rgw_utils
 
 from teuthology import misc as teuthology
-from util.rgw import rgwadmin
+from util.rgw import rgwadmin, get_user_summary, get_user_successful_ops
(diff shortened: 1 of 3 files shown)


dachary:
The same four jobs were scheduled, with and without this patch:

* http://pulpito.ceph.com/loic-2015-05-08_12:17:16-rgw-firefly-backports---basic-multi with this patch  passes works
* http://pulpito.ceph.com/loic-2015-05-08_12:16:53-rgw-firefly-backports---basic-multi without this patch fails

PR #440, opened 2015-05-13
Reduce fs test node usage

gregsfortytwo:
This reconfigures the cluster yaml fragments for most of the filesystem-related tests to reduce the number of nodes they use.


gregsfortytwo:
I removed the kclient stuff from this PR and will get it merged separately. The patches here are slightly simplified from the revert and merge stuff used in http://pulpito.ceph.com/gregf-2015-06-03_14:53:22-fs-greg-fs-testing-3---basic-multi/ (against testing commit d3482fd118d050455e9b991309ea37c030581a69).

jcsp:
All looks sane to me :+1:

PR #441, opened 2015-05-13
Removed qemu_iozone.yaml.disabled as it broke teuthology-suite #11605

yuriw:
@dmick @jdurgin Pls review


dmick:
I think leaving the file there is fine; rather, teuthology-suite should not be so picky.

gregsfortytwo:
This is fixed now in teuthology-suite, right?

yuriw:
@gregsfortytwo yes, removing it

PR #442, opened 2015-05-13
cephfs/test_client_limits.py: invalidate kernel dcache according to c…

ukernel:
…lient_cache_size

### DIFF
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -340,3 +340,13 @@ print find_socket("{client_name}")
         """
         status = self._admin_socket(['status'])
         return status['osd_epoch'], status['osd_epoch_barrier']
+
+    def get_dentry_count(self):
+        """
+        Return 2-tuple of dentry_count, dentry_pinned_count
+        """
+        status = self._admin_socket(['status'])
+        return status['dentry_count'], status['dentry_pinned_count']
+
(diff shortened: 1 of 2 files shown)


jcsp:
:+1: looks great, should merge as soon as https://github.com/ceph/ceph/pull/4653

gregsfortytwo:
This looks good as far as it goes. But like I said in https://github.com/ceph/ceph/pull/4653, I'd really like to see this including checks on the MDS's cache pressure requests. Is that infeasible for some reason?

ukernel:
On Thu, May 14, 2015 at 5:23 AM, Gregory Farnum <notifications@github.com>
wrote:

> This looks good as far as it goes. But like I said in ceph/ceph#4653
> <https://github.com/ceph/ceph/pull/4653>, I'd really like to see this
> including checks on the MDS's cache pressure requests. Is that infeasible
> for some reason?
>
there already is test case for MDS's cache pressure. I don't like the idea
that testing multiple functions in single test case.

jcsp:
@gregsfortytwo the MDS-driven cache pressure is orthogonal to this (it's the Client::trim_caps path, this is about the Client::trim_cache path)

gregsfortytwo:
Oh, I guess we are doing the kernel invalidate in trim_caps already.

gregsfortytwo:
Running in greg-fs-testing as a final check.

PR #444, opened 2015-05-26
tasks/cephfs: mount fusectl before listing fuse connections

ukernel:
Fixes: #11756

### DIFF
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -68,10 +68,17 @@ class FuseMount(CephFSMount):
         run_cmd.extend(run_cmd_tail)
 
         def list_connections():
+            self.client_remote.run(
+                args=["sudo", "mount", "-t", "fusectl", "/sys/fs/fuse/connections", "/sys/fs/fuse/connections"],
+                check_status=False
+            )
             p = self.client_remote.run(
                 args=["ls", "/sys/fs/fuse/connections"],
-                stdout=StringIO()
+                stdout=StringIO(),
+                check_status=False
             )
+            if p.exitstatus != 0:
+                return []
 
             ls_str = p.stdout.getvalue().strip()
             if ls_str:


tasks/cephfs/fuse_mount.py
```diff
@@ -68,6 +68,13 @@ def mount(self):
         run_cmd.extend(run_cmd_tail)
 
         def list_connections():
+            self.client_remote.run(
```
jcsp: Would it be correct to simply return [] if the directory is not there?  We shouldn't have to manually modprobe fuse

ukernel:
updated

tasks/cephfs/fuse_mount.py
jcsp: Do we still even need the mount call here?  I don't know the details but I always assumed this was something that fuse was supposed to take care of for us, and if it isn't present then we can assume there are no fuse connections.

If it is necessary for some reason, then the right place to call this is probably at the start of mount() rather than in each list_connections()

ukernel: In my test envirorment, fusectl does not get mounted after mounting cephfs

ukernel: On Tue, May 26, 2015 at 8:00 PM, John Spray <notifications@github.com>
wrote:

> In tasks/cephfs/fuse_mount.py
> <https://github.com/ceph/ceph-qa-suite/pull/444#discussion_r31025977>:
>
> > @@ -68,6 +68,13 @@ def mount(self):
> >          run_cmd.extend(run_cmd_tail)
> >
> >          def list_connections():
> > +            self.client_remote.run(
>
> Do we still even need the mount call here? I don't know the details but I
> always assumed this was something that fuse was supposed to take care of
> for us, and if it isn't present then we can assume there are no fuse
> connections.
>
> If it is necessary for some reason, then the right place to call this is
> probably at the start of mount() rather than in each list_connections()
>
if we mount fusectl at the start of mount, It's possible that the kernel
still does not have fuse.ko inserted.

PR #452, opened 2015-05-31
Logrotate

gregsfortytwo:
This PR currently includes log rotation in the Ceph task that you can configure via yaml. It works in the small set of tests I've run against it, but is not yet enabled anywhere by default and shouldn't be merged.

I'm testing it against an FS run and will update the PR with config changes for that once it passes.

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -94,35 +94,52 @@ def ceph_log(ctx, config):
             self.stopping = True
             self.thread.get()
             
-    def write_rotate_conf(ctx):
+    def write_rotate_conf(ctx, daemons):
         testdir = teuthology.get_testdir(ctx)
         rotate_conf_path = os.path.join(os.path.dirname(__file__), 'logrotate.conf')
-        conf = file(rotate_conf_path, 'rb').read() # does this leak an fd or anything?
-        for remote in ctx.cluster.remotes.iterkeys():
-            teuthology.write_file(remote=remote,
-                                  path='{tdir}/logrotate.mds.conf'.format(tdir=testdir),
-                                  data=conf
-                              )
-            remote.run(
-                args=[
-                    'sudo',
-                    'mv',
-                    '{tdir}/logrotate.mds.conf'.format(tdir=testdir),
-                    '/etc/logrotate.d/ceph-test.conf'
-                ]
-            )
+        with file(rotate_conf_path, 'rb') as f:
+            conf = ""
+            for daemon,size in daemons.iteritems():
+                log.info('writing logrotate stanza for {daemon}'.format(daemon=daemon))
+                conf += f.read().format(daemon_type=daemon,max_size=size)
+                f.seek(0, 0)
+            
+            for remote in ctx.cluster.remotes.iterkeys():
+                teuthology.write_file(remote=remote,
+                                      path='{tdir}/logrotate.ceph-test.conf'.format(tdir=testdir),
+                                      data=StringIO(conf)
+                                  )
+                remote.run(
+                    args=[
+                        'sudo',
+                        'mv',
+                        '{tdir}/logrotate.ceph-test.conf'.format(tdir=testdir),
+                        '/etc/logrotate.d/ceph-test.conf',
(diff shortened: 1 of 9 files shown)


gregsfortytwo:
http://pulpito.ceph.com/gregf-2015-06-03_14:53:22-fs-greg-fs-testing-3---basic-multi/ (against qa-suite commit d3482fd118d050455e9b991309ea37c030581a69)

gregsfortytwo:
This PR depends on #440 and was tested at the same time. We may want to dial back the log rotation if we find it annoying, since we don't need it for the typica nodes now, but it might as well go in-tree.

jcsp:
Oops, left a couple of comments on a commit instead of the overall files changed diff (https://github.com/ceph/ceph-qa-suite/commit/5935f86e49df1f2a13a9c90daa8a0ea2d2754871)

tasks/ceph.py
```diff
@@ -71,10 +72,79 @@ def ceph_log(ctx, config):
             )
         )
 
+    class Rotater:
```
jcsp: Very minor style thing: new python code uses "class Rotater(object):" in order to be a "new style" class.  Doesn't actually matter unless you're doing inheritance etc.

tasks/ceph.py
```diff
@@ -71,10 +72,79 @@ def ceph_log(ctx, config):
...
+                        '/etc/logrotate.d/ceph-test.conf',
+                        run.Raw('&&'),
+                        'sudo',
+                        'chown',
+                        'root.root',
+                        '/etc/logrotate.d/ceph-test.conf'
+                    ]
+                )
+
+    if ctx.config.get('log-rotate'):
```
jcsp: Since this setting is being acted upon by a particular task, it would probably be cleaner to put it inside the ceph task's config dict (and then set it with overrides in .yaml files), rather than a new toplevel item.  But it doesn't matter much because the users are all within ceph-qa-suite too so it's free to change the definition of this at will in the future if we feel like cleaning it up.

jcsp:
Looks fine.  Only comment of substance is about using a gevent.event.Event() for Rotator.stopping, so that we don't insert a 0-30s delay on shutdown

tasks/ceph.py
gregsfortytwo: Mmm, yeah. TBH I just did that because I got frustrated with getting all the nested structures right (*blush*) but I also like it because we might want to rotate non-Ceph stuff in the future. Right now it's part of ceph.py but given the structure we can move it elsewhere pretty easily.

gregsfortytwo:
Good?

ktdreyer:
Can we stop rm'ing `/etc/logrotate.d/ceph` and write to that file instead?

It would avoid the need for hacks like https://github.com/ceph/ceph-qa-suite/pull/520

gregsfortytwo:
@ktdreyer, we rm the default logrotate because the way it works was periodically breaking our log gathering at the end of jobs — we need the log files to not change unexpectedly during our testing, since we might be grabbing them for archival.
So this PR, when we decided we needed to do log rotation, is creating temporary logrotate config files and invoking them in a controlled fashion; I think it's doing so idiomatically but I don't really know. In any case if we want to use the ones in the packaging they'll need to become significantly more complicated and configurable than the packaged ones are right now. I don't know if that's appropriate just to enable some test debugging?

ktdreyer:
I realize my original message may have been unclear.

I don't care if we zero out the package's /etc/logrotate.d/ceph and put in our own contents. That's fine. It's even close to what an (advanced) user might do.

The issue is that we're writing an entirely new file to /tmp, and then mv'ing it into place. The file ends up with an entirely different selinux context, which requires other hacks like setting the selinux context explicitly on this file after it's been moved into place, etc.

So it's not so much about using the package's logrotate file *contents* - I get that teuthology needs an entirely separate schedule for log rotation, etc. It's about using the same filename and inode so we preserve the selinux xattrs.

gregsfortytwo:
Ah. PRs welcome. :)

PR #453, opened 2015-06-03
tasks/cephfs: time out on ceph-fuses that don't die

jcsp:
For cases where we have e.g. poked the fuse abort
file for a process, but it's still not dying.  Because
this is a special class of error (unlike e.g. when
we force umount something because the network is gone)
raise the error instead of trying again to kill
the client.

Fixes: #11835

### DIFF
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -7,6 +7,7 @@ import logging
 from textwrap import dedent
 
 from teuthology import misc
+from teuthology.contextutil import MaxWhileTries
 from teuthology.orchestra import run
 from teuthology.orchestra.run import CommandFailedError
 from .mount import CephFSMount
@@ -234,7 +235,12 @@ class FuseMount(CephFSMount):
 
         try:
             if self.fuse_daemon:
-                self.fuse_daemon.wait()
+                # Permit a timeout, so that we do not block forever
+                run.wait([self.fuse_daemon], 30)
+        except MaxWhileTries:
+            log.error("process failed to terminate after unmount.  This probably"
+                      "indicates a bug within ceph-fuse.")
+            raise
         except CommandFailedError:
             pass
 
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -236,7 +236,7 @@ class FuseMount(CephFSMount):
         try:
             if self.fuse_daemon:
                 # Permit a timeout, so that we do not block forever
-                run.wait([self.fuse_daemon], 30)
+                run.wait([self.fuse_daemon], 900)
         except MaxWhileTries:
             log.error("process failed to terminate after unmount.  This probably"
                       "indicates a bug within ceph-fuse.")


tasks/cephfs/fuse_mount.py
```diff
@@ -234,7 +235,12 @@ def umount_wait(self, force=False):
 
         try:
             if self.fuse_daemon:
-                self.fuse_daemon.wait()
+                # Permit a timeout, so that we do not block forever
+                run.wait(self.fuse_daemon, 30)
+        except MaxWhileTries:
+            log.error("process failed to terminate after unmount.  This probably"
+                      "indicates a bug within ceph-fuse.")
+            raise
```
gregsfortytwo: What does this actually do to the teuthology test runs that hit it? I've no idea but I think most throws in the post-yield cleanup result in hung jobs and nothing getting cleaned up...

jcsp: @gregsfortytwo in run_tasks.py, in the part following the "Unwinding manager..." log message, it looks to me like exceptions in teardown are explicitly handled.

Entirely possible for a rogue process to be left lying around, but that's why we have nuke

gregsfortytwo:
Looks like this is busting things up; several failures in ubuntu-2015-06-09_11:25:56-fs-greg-fs-testing---basic-multi.

>2015-06-09T12:18:44.892 INFO:teuthology.orchestra.run.plana91:Running: 'sudo fusermount -u /home/ubuntu/cephtest/mnt.0'
2015-06-09T12:18:45.017 INFO:tasks.cephfs.fuse_mount.ceph-fuse.0.plana91.stderr:ceph-fuse[19072]: fuse finished with error 0 and tester_r 0
2015-06-09T12:18:45.041 INFO:teuthology.orchestra.run.plana91:Running: "stat --file-system '--printf=%T\n' -- /home/ubuntu/cephtest/mnt.0"
2015-06-09T12:18:45.050 DEBUG:tasks.cephfs.fuse_mount:ceph-fuse not mounted, got fs type 'ext2/ext3'
2015-06-09T12:18:45.050 ERROR:teuthology.run_tasks:Manager failed: ceph-fuse
Traceback (most recent call last):
  File "/home/teuthworker/src/teuthology_master/teuthology/run_tasks.py", line 125, in run_tasks
    suppress = manager.__exit__(*exc_info)
  File "/usr/lib/python2.7/contextlib.py", line 24, in __exit__
    self.gen.next()
  File "/var/lib/teuthworker/src/ceph-qa-suite_greg-fs-testing/tasks/ceph_fuse.py", line 135, in task
    mount.umount_wait()
  File "/var/lib/teuthworker/src/ceph-qa-suite_greg-fs-testing/tasks/cephfs/fuse_mount.py", line 239, in umount_wait
    run.wait(self.fuse_daemon, 30)
  File "/home/teuthworker/src/teuthology_master/teuthology/orchestra/run.py", line 393, in wait
    not_ready = list(processes)
TypeError: 'RemoteProcess' object is not iterable

jcsp:
Oops, that was a typo to run.wait() (it wants a list of processes).  Updated.

PR #454, opened 2015-06-05
Single node ceph-deploy tests to exercise commonly used cli

vasukulkarni:
Single Node Ceph-deploy tests to exercise commonly used CLI and test basic init startup
  - Need to add 1 node yaml for this test in smoke suite

### DIFF
--- a/tasks/ceph_deploy.py
+++ b/tasks/ceph_deploy.py
@@ -436,6 +436,150 @@ def build_ceph_cluster(ctx, config):
         execute_ceph_deploy(purgedata_nodes)
 
 
+def execute_cdeploy(admin,cmd,path):
+    """Execute ceph-deploy commands """
+    """Either use git path or repo path """
+    if path is not None:
+       ec= admin.run(
+            args=[
+                 'cd',
+                 run.Raw('~/cdtest'),
+                 run.Raw(';'),
+                '{path}/ceph-deploy/ceph-deploy'.format(path=path),
+                 run.Raw(cmd),
+                ],
+               check_status=False,
+           ).exitstatus
+    else:
+      ec= admin.run(
+              args=[
+                  'cd',
+                   run.Raw('~/cdtest'),
+                   run.Raw(';'),
+                  'ceph-deploy',
+                  run.Raw(cmd),
+                ],
+              check_status=False,
+              ).exitstatus
+    if ec != 0:
+       raise RuntimeError ("failed during ceph-deploy cmd: {cmd} , ec={ec}".format(cmd=cmd,ec=ec)) 
+
+@contextlib.contextmanager
+def cli_test(ctx, config):
+    """
+     ceph-deploy cli to exercise most commonly use cli's and ensure
+     all commands works and also startup the init system.
+        
+    """
+    log.info('Ceph-deploy Test')
+    if config is None:
+        config = {}
+        
+    if config.get('rhbuild'):
+        path=None
+    else:
+        path = teuthology.get_testdir(ctx)  
+    mons = ctx.cluster.only(teuthology.is_type('mon'))
+    for node,role in mons.remotes.iteritems():
+        admin=node
+        admin.run( args=[ 'mkdir', '~/', 'cdtest' ],check_status=False)
+        nodename=admin.shortname
+    system_type = teuthology.get_system_type(admin)
+    if config.get('rhbuild'):
+        admin.run(args = ['sudo', 'yum', 'install', 'ceph-deploy', '-y'])
+    log.info('system type is %s', system_type)
+    osds = ctx.cluster.only(teuthology.is_type('osd'))
+   
+    for remote,roles in osds.remotes.iteritems():
+        devs = teuthology.get_scratch_devices(remote)
+        log.info("roles %s" , roles)
+        if (len(devs) < 3):
+            log.error('Test needs minimum of 3 devices, only found %s', str(devs))
+            raise RuntimeError ( "Needs minimum of 3 devices ")
+    
+    new_cmd= 'new ' + nodename
+    new_mon_install = 'install --mon ' + nodename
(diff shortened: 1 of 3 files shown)


tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+    num_of_disk=0
+    for i in range(3):
+        zap_disk = 'disk zap '  + "{n}:{d}".format(n=nodename,d=devs[i])
+        prepare= 'osd prepare ' + "{n}:{d}".format(n=nodename,d=devs[i])
+        execute_cdeploy(admin,zap_disk,path)
+        execute_cdeploy(admin,prepare,path)
+        
+    admin.run(args=['ls',run.Raw('-lt'),run.Raw('~/cdtest/')])
+    time.sleep(4)
+    r = remote.run(args=['sudo', 'ceph','-s'],check_status=False)
```
trhoden: since `r` is not used here, and re-assigned in the next line, what does this command do?

tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+        nodename=admin.shortname
+    system_type = teuthology.get_system_type(admin)
+    if config.get('rhbuild'):
+        admin.run(args = ['sudo', 'yum', 'install', 'ceph-deploy', '-y'])
+    log.info('system type is %s', system_type)
+    osds = ctx.cluster.only(teuthology.is_type('osd'))
+    for remote,roles in osds.remotes.iteritems():
+        devs = teuthology.get_scratch_devices(remote)
+        log.info("roles %s" , roles)
+        if (len(devs) < 3):
```
trhoden: I know this is mean to be a single-node test, but do you want to require 3 OSDs?  It may limit your choice of hardware for the smoke tests.  Might be better to go for 2 (and run tests with pool size 2).  I'm not entirely sure on that one though, as I don't all the background here.

tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+            log.error('Test needs minimum of 3 devices, only found %s', str(devs))
+            raise RuntimeError ( "Needs minimum of 3 devices ")
+    new_cmd= 'new ' + nodename
+    new_mon_install = 'install --mon ' + nodename
+    new_osd_install = 'install --osd ' + nodename
+    new_admin = 'install --cli ' + nodename
```
trhoden: since this is a single-node tests, adding ``--cli`` isn't necessary, as the CLI will get installed when either mon or OSD is installed (and you are doing both).

tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+    create_initial= '--overwrite-conf mon create-initial ' + nodename
+    execute_cdeploy(admin,new_cmd,path)
+    execute_cdeploy(admin,new_mon_install,path)
+    execute_cdeploy(admin,new_osd_install,path)
+    execute_cdeploy(admin,new_admin,path)
+    execute_cdeploy(admin,create_initial,path)
+    num_of_disk=0
```
trhoden: should this be ``len(devs)`` ?

tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+    for i in range(3):
```
trhoden: and then iterate over ``num_of_disk instead``?

Or maybe do `` for i, device in enumerate(devs)`` instead, and not need ``num_of_disk`` at all?

tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+    finally:
+        log.info("cleaning up")
+        if system_type == 'deb':
+            remote.run(args=['sudo', 'stop','ceph-all'],check_status=False)
+            remote.run(args=['sudo', 'service','ceph', '-a', 'stop'],check_status=False)
+        else:
+            remote.run(args=['sudo', '/etc/init.d/ceph', '-a', 'stop'],check_status=False)
+        time.sleep(4)
+        num_of_disk=0
+        for i in range(3):
```
trhoden: what ever you do with ``num_of_disk`` above, be sure to account for here.

tasks/ceph_deploy.py
```diff
@@ -436,6 +436,140 @@ def execute_ceph_deploy(cmd):
...
+            umount_dev = "{d}1".format(d=devs[i])
+            r = remote.run(args=['sudo', 'umount',run.Raw(umount_dev)])
+            num_of_disk=num_of_disk+1
```
trhoden: and this doesn't seem to be used for anything.

tasks/ceph_deploy.py
vasukulkarni: @trhoden  good catch, i dont need num_of_disk, it was old logic forgot to remove it  iteration overs devs is sufficient.

tasks/ceph_deploy.py
vasukulkarni: @trhoden will remove this variable

tasks/ceph_deploy.py
vasukulkarni: @trhoden I just added this to exercise the cli, I knew this will get installed before but just rerunning this to cover the cli code path.

tasks/ceph_deploy.py
vasukulkarni: @trhoden , will remove the first 'r', its not used, just need the output of ceph -s in logs.

trhoden: Ah, that makes sense!  I didn't consider running commands just to see output in the logs.  :)

trhoden:
This looks good to me. :+1:

PR #457, opened 2015-06-11
tasks/cephfs: fix race in test_full

jcsp:
Sometimes mount A would get a cap revoke when mount
B did its last IO, resulting in mount A's OSD epoch
getting updated too.

Fix by making sure mount B is the last one to have
done IO before we do the barrier, so that when
it does IO again after the barrier, mount A can't
be holding any caps that B would need.

Fixes: #11913


ukernel:
make sense for me

gregsfortytwo:
We don't have enough PRs for a proper greg-fs-testing so I'm just going to merge this and if it breaks we'll fix it.

PR #460, opened 2015-06-12
ceph: fix up log rotation stopper

gregsfortytwo:


jcsp:
should be Fixes: #11986

PR #461, opened 2015-06-16
cephfs/test_client_limits.py: update test_client_oldest_tid

ukernel:
merge this after https://github.com/ceph/ceph/pull/4791


gregsfortytwo:
http://pulpito.ceph.com/ubuntu-2015-09-22_15:56:36-fs-greg-fs-testing---basic-multi/1064720/ illustrates this testing not behaving properly.
(The client did send along its updates correctly, so I merged the actual branch itself.).

PR #464, opened 2015-06-17
Tests for damage handling and cephfs-data-scan

jcsp:


jcsp:
NB Will also needs updates to log whitelist

gregsfortytwo:
The wip-damaged-fixes branch got merged to master a couple weeks ago, so this should get reviewed and merged, right?
Does it have those log whitelists handled?

gregsfortytwo:
Whoops, I guess this is for the still-pending bits.

PR #465, opened 2015-06-17
hammer suite split - older and newer

yuriw:
Fixes #11570

Split suite /upgrade/hammer/ in two parts "older" and "newer" and removed rgw tests from the "older" part.


zmc:
I'd like to see a more informative commit message

PR #466, opened 2015-06-18
Tests for wip-damaged-fixes

jcsp:
This stuff is also in wip-offline-backward but separated out here.

DNM until wip-damaged-fixes in ceph tree


jcsp:
@gregsfortytwo the related parts are merged in ceph tree now (these tests pass)

jcsp:
@ukernel ping, this is ready for review since the ceph parts merged

PR #471, opened 2015-06-26
Per Yehuda's comment in #12154, set s3tests branch to appropriate assignment  

yuriw:
@yehudasa Pls take a look and it OK I will cherry pink on hammer

### DIFF
--- a/suites/upgrade/hammer/basic/2-workload/s3tests.yaml
+++ b/suites/upgrade/hammer/basic/2-workload/s3tests.yaml
@@ -4,6 +4,6 @@ workload:
   - print: "**** done rgw: [client.0] 2-workload"
   - s3tests:
       client.0:
-        force-branch: firefly-original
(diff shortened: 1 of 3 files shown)


yehudasa:
@yuriw I don't think this is right. If upgrading from hammer to hammer it should be using the *hammer* branch. Beforehand it was supposed to use the firefly-original branch, but it was actually using the master branch. That was wrong. Something is broken so that it ignores that branch completely.

yuriw:
@yehudasa see changes, pls

yehudasa:
@yuriw I don't know if it's going to work or not. You'll have to run it to verify that it's actually doing the correct thing.

yuriw:
@yehudasa ok running now http://pulpito.ceph.com/teuthology-2015-06-26_13:04:00-upgrade:hammer-hammer---basic-vps/

PR #472, opened 2015-06-27
Add a test case to turn on the feature flag to read all replicas for EC pool

guangyy:
This test is to cover the following feature:
   osd_pool_erasure_code_subread_all

Signed-off-by: Guang Yang <yguang@yahoo-inc.com>


guangyy:
@athanatos

guangyy:
Close this one, and open a new one according to the latest discussion: https://github.com/ceph/ceph-qa-suite/pull/499

PR #475, opened 2015-06-29
calamari_setup.py: Missing args= keyword in umount call

dmick:


dmick:
wusui already filed one.

PR #478, opened 2015-07-01
cephfs/test_auto_repair.py: flush journal after umount

jcsp:
When client capabilities get released, MDS may update corresponding
inodes' client writable range and mark those inodes dirty. The auto
repair test expects MDS to trim inodes from its cache, but MDS can't
trim dirty inodes. So we should flush journal after umount.

Fixes: #12172
(cherry picked from commit 835ad8b8f3038290846c612b66f3bffcc5755338)


jcsp:
You might be thinking "Hey, didn't we just merge this?"... and you'd be right... to master, because I foolishly opened the previous PR against the wrong branch.  Doh.

This time targeting next.

PR #479, opened 2015-07-01
rgw: do not enable both tcp and uds for fastcgi

andrewschoen:
This fixes a bug where we were setting up mod_fastcgi to use both tcp
and uds.


athanatos:
Worked when I tested it.

PR #486, opened 2015-07-11
increase fuse mount timeout to 300 seconds

dachary:
The 30 seconds value is too conservative when running on virtual
machines because they tend to be slow. The timeout is useful to avoid
blocking forever, but it can be as long as 5 minutes without creating
any actual problem.

### DIFF
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -97,7 +97,7 @@ class FuseMount(CephFSMount):
         while list_connections() == pre_mount_conns:
             time.sleep(1)
             waited += 1
-            if waited > 30:
+            if waited > 300:
                 raise RuntimeError("Fuse mount failed to populate /sys/ after {0} seconds".format(
                     waited
                 ))


jcsp:
It's not a good idea to run on pathologically slow machines.  There are timeouts elsewhere in the system that you will hit, like the client session eviction.  There are also tests that are deliberately sensitive to timing (they test rate limiting and timeouts).

jcsp:
I think I'd actually rather the tests failed nice and early during mount than in more mysterious ways later on

dachary:
This timeout creates false negative, always, when running on OpenStack provided virtual machines. I've also seen this error show up occasionally on bare metal. Maybe 300 seconds is too much. But it looks like 30 seconds is too conservative.

gregsfortytwo:
Is it just a matter of waiting for the Ceph cluster to be fully set up, or is the actual mount process taking so long?

Either way, perhaps we need a separate set of timeouts for VM test clusters. :/

dachary:
https://github.com/ceph/ceph-qa-suite/pull/482/files#diff-d9ed181636bd9d3aa5796ef1e1589bd1R101 is what fails and I'm not sure exactly what is causing it.

It ran successfully a four or five time this week-end after increasing the timeout (I was re-running the upgrade suite of hammer over and over, running into other issues such as not enough memory or not enough disk space).

There are VM specific timeouts at https://github.com/ceph/ceph-qa-suite/blob/master/machine_types/vps.yaml already.

gregsfortytwo:
Right, so there are two things we can do here:

can give them 60 seconds or something.
for VM tests.

dachary:
@gregsfortytwo I'll amend this pull request accordingly. @gregsfortytwo @jcsp thanks for the review :-)

jcsp:
OK, we can see how we go, but let's make sure any failures from VM runs are marked as such so that we pick up when problems are timing related.

dachary:
fix in master instead of hammer https://github.com/ceph/ceph-qa-suite/pull/489

PR #487, opened 2015-07-14
calamari_setup: handle new structure in 1.3.0 (MON and OSD repos)

dmick:
Also, clean up the command handling; failure will throw an exception,
so no need to accumulate the errors in an int

Fixes: #12228


ktdreyer:
Thanks for the comments and getting rid of the `for entry in first_cmds` loop. This makes it easier to quickly read and understand what's going on.

PR #489, opened 2015-07-14
tasks/ceph_fuse.py: virtual machines need flexible mount timeout

dachary:
can give them 60 seconds.
for VM tests.

http://tracker.ceph.com/issues/12320 Fixes: #12320

### DIFF
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -100,11 +100,16 @@ class FuseMount(CephFSMount):
         self.fuse_daemon = proc
 
         # Wait for the connection reference to appear in /sys
+        mount_wait = self.client_config.get('mount_wait', 0)
+        if mount_wait > 0:
+            log.info("Fuse mount waits {0} seconds before checking /sys/".format(mount_wait))
+            time.sleep(mount_wait)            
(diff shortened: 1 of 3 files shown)


tasks/cephfs/fuse_mount.py
```diff
@@ -100,11 +100,16 @@ def list_connections():
         self.fuse_daemon = proc
 
         # Wait for the connection reference to appear in /sys
+        mount_wait = self.client_config.get('mount_wait', 0)
+        if mount_wait > 0:
+            log.info("Fuse mount waits {0} seconds before checking /sys/")
+            time.sleep(mount_wait)
```
jcsp: Hmm, why do we sleep first and then start checking?  Why not use a 180s timeout instead of a 60s wait and a 120s timeout?

Other than that this looks good.

dachary: I did it this way because @gregsfortytwo suggested "add a wait time before the mount attempt to let the cluster get set up.". Did I get this wrong ? I'm not opposed to a 180s timeout, just following suggestions.

dachary:
@jcsp I'm running a suite to verify the change does the right thing

dachary:
@jcsp it passes a test run as shown at http://tracker.ceph.com/issues/12320#note-1

PR #491, opened 2015-07-14
DNM: fs resource requirement hints

dachary:
http://tracker.ceph.com/issues/12329

### DIFF
--- a/suites/fs/verify/tasks/cfuse_workunit_suites_dbench.yaml
+++ b/suites/fs/verify/tasks/cfuse_workunit_suites_dbench.yaml
@@ -10,3 +10,11 @@ tasks:
     clients:
       all:
         - suites/dbench.sh
+openstack:
+  machine:
+    disk: 100 # GB
+    ram: 4096 # MB
+    cpus: 1
+  volumes: # attached to each instance
+    count: 0
+    size: 1 # GB


gregsfortytwo:
I'm confused, is this a hint pattern you're following or something? Are we going to need to reorganize every single test with resource usage into its own folder? :/

dachary:
@gregsfortytwo modified the yaml file instead

jcsp:
I'm not violently opposed, but it seems like disk/cpu/mem requirements aren't really an openstack specific thing, and it feels a bit weird to have an "openstack" section in an individual filesystem test.  Maybe these could be more generic, and also enforced on physical hosts by having the test fail out if it tries to run on a physical host that doesn't match requirements?

dachary:
@jcsp the thing is I have no clue how to make that actually generic (provisioning hardware is way too complicated for me ;-). And the history of ceph-qa-suite teuthology has a lot of seemingly generic things that turn out to be specialized in strange and confusing ways. Hence the idea of having something very specific so as to not encourage that type of drift. As of now and probably for the year to come and more, nothing else but OpenStack will interpret these anyway.

gregsfortytwo:
We can make this generic later if it looks to be useful elsewhere.

gregsfortytwo:
@dachary I assume you actually do want this to get merged? (See: DNM in title)

dachary:
@gregsfortytwo not just yet. there are a few more that needs to be added.

dachary:
the branch does not carry a descriptive name and the tests really are against hammer, not master https://github.com/ceph/ceph-qa-suite/pull/497

dachary:
in favor of https://github.com/ceph/ceph-qa-suite/pull/497

PR #493, opened 2015-07-14
Per IRC chat by Yehuda, Josh and Dan:

yuriw:
"the purpose of firefly-original in general was for upgrades from prior point releases of firefly, where newer tests on the firefly branch would fail
it shouldn't be used if it's just a firefly->hammer upgrade though, the plain old firefly branch should be fine for that"

Changed s3tests ref to firefly


yuriw:
bad commit

PR #499, opened 2015-07-21
rados: add a test case to thrash the feature of fast read on EC pools

guangyy:
Try to test the PR - https://github.com/ceph/ceph/pull/5309

Signed-off-by: Guang Yang yguang@yahoo-inc.com


dzafman:
Merged as #503

PR #505, opened 2015-07-23
calamari_setup: disable epel again after ceph-deploy

dmick:
ceph-deploy enables EPEL on Centos; this caused problems for
Calamari.  Undo it after ceph-deploy has installed Ceph.

### DIFF
--- a/tasks/calamari_setup.py
+++ b/tasks/calamari_setup.py
@@ -65,6 +65,8 @@ def task(ctx, config):
         lambda: adjust_yum_repos(ctx, cal_svr, config['no_epel']),
         lambda: calamari_install(config, cal_svr),
         lambda: ceph_install(ctx, cal_svr),
+        # do it again because ceph-deploy installed epel for centos
+        lambda: remove_epel(ctx, config['no_epel']),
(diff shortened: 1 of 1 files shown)


tasks/calamari_setup.py
```diff
@@ -153,6 +155,24 @@ def fix_yum_repos(remote, distro):
...
+def remove_epel(ctx, no_epel):
+    """
+    just remove epel.  No undo; assumed that it's used after
+    adjust_yum_repos, and relies on its state-save/restore.
+    """
+    if no_epel:
+        for remote in ctx.cluster.remotes:
+            if remote.os.name.startswith('centos'):
+                remote.run(args= [
+                    'sudo', 'rm', '-f', run.Raw('/etc/yum.repos.d/epel*')
```
trhoden: would you like to do it this way, or something more like `yum-config-manager --disable epel`.  That assumes that `yum-utils` is installed, though.

dmick: assuming yum-utils is there is why it doesn't; there's older code here and in calamari_nosetests that just manipulates the presence of the repo files.  It's brutish, but effective.

PR #509, opened 2015-07-24
Fix for wip-12225, Extra argument to mon create-initial was never …

vasukulkarni:
off_rhoden: this can wait for 12081 issue and can be merged after that is verfied.


trhoden:
Roger.  I'll let you and Yuri work out when best to merged, but the fix looks good and this gets a :+1: from me.  Thanks!

PR #510, opened 2015-07-25
calamari_setup: Add --no-adjust-repos to ceph-deploy install --<type>

dmick:
Otherwise, ceph-deploy will install an apt source that points to
ceph.com, which will override the local ISO repos.

Also, fix a cut'n'paste bug (the second set of installs needs to be
--osd).  This was harmless, as the ceph-deploy currently installs
everything, but would have broken the once package splits happen.

### DIFF
--- a/tasks/calamari_setup.py
+++ b/tasks/calamari_setup.py
@@ -109,6 +109,11 @@ def fix_yum_repos(remote, distro):
     contain a repo file named rhel<version-number>.repo
     """
     if distro.startswith('centos'):
+        # hack alert: detour: install lttng for ceph
+        # this works because epel is preinstalled on the vpms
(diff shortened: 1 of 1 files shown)


trhoden:
Looks right.

As you've seen, this will only work with ceph-deploy >= 1.5.26, otherwise the use of `--no-adjust-repos` with `--mon` or `--osd` will throw an error. :/

If you are testing with 1.5.25 we will want to account for that.

dmick:
Yeah, I now know we want another change here.

dmick:
@trhoden: updated with more discovered gunge.  it seems to pass now; see what you think.

trhoden:
This all looks good to me.

PR #513, opened 2015-07-28
vstart_runner for CephFS

jcsp:


jcsp:
As introduced on mailing list: http://thread.gmane.org/gmane.comp.file-systems.ceph.devel/26151

Rather than split up the client handling into a part that operates as root, I've just added enough infrastructure to skip over the tests that require client trimming to work, and use new mds_root_ino_uid settings to avoid the need for root when operating within filesystem mounts.

tasks/cephfs/filesystem.py
```diff
@@ -73,16 +73,38 @@ def get_pgs_per_fs_pool(self):
...
+            self.mon_manager.raw_cluster_cmd('fs', 'rm', fs_data['name'], '--yes-i-really-mean-it')
+            self.mon_manager.raw_cluster_cmd('osd', 'pool', 'delete',
+                                             fs_data['metadata_pool'], fs_data['metadata_pool'],
+                                             '--yes-i-really-really-mean-it')
+            for data_pool in fs_data['data_pools']:
+                self.mon_manager.raw_cluster_cmd('osd', 'pool', 'delete',
+                                                 data_pool, data_pool,
+                                                 '--yes-i-really-really-mean-it')
 
     def delete(self):
```
gregsfortytwo: I don't think anybody else invokes delete(); we should probably just delete it so it doesn't get outdated and then pick up an unaware user.

jcsp: agreed -- updated.

gregsfortytwo:
I'm sure I'm just missing something but I'm not sure how we're actually supposed to run this. Invoke vstart and then invoke vstart_runner from inside the ceph src directory? :)

jcsp:
Yeah, pretty much: the magic part is setting PYTHONPATH correctly, here's an example command line:
```
PYTHONPATH=~/git/teuthology/:~/git/ceph-qa-suite/ python ~/git/ceph-qa-suite/tasks/cephfs/vstart_runner.py tasks.cephfs.test_data_scan.TestDataScan.test_stashed_layout
```

jcsp:
I've just added some handy warning messages for when someone tries to run this without the right paths.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,782 @@
...
+        subproc = subprocess.Popen(args,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.PIPE,
+                                   stdin=subprocess.PIPE,
+                                   cwd=cwd)
+
+        if stdin and isinstance(stdin, basestring):
+            # Hack: writing to stdin is not deadlock-safe, but it "always" works
+            # as long as the input buffer is "small"
+            subproc.stdin.write(stdin)
```
gregsfortytwo: Hmm, it looks like this just throws away the stdin param if it's not a string? I'm not sure we even expect it to be a string here, but maybe I've missed something, and surely we shouldn't silently ignore it.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,782 @@
...
+        )
+
+        if wait:
+            proc.wait()
+
+        return proc
+
+
+# FIXME: twiddling vstart daemons is likely to be unreliable, we should probably just let vstart
+# run RADOS and run the MDS daemons directly from the test runner
```
gregsfortytwo: Can you discuss this a bit more? In particular I notice that this is definitely manipulating daemons but doesn't update the pid files vstart uses — how does a tested cluster get shut down properly?

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,782 @@
...
+
+        return None
+
+    def wait(self, timeout):
+        waited = 0
+        while self._get_pid() is not None:
+            if waited > timeout:
+                raise MaxWhileTries("Timed out waiting for daemon {0}.{1}".format(self.daemon_type, self.daemon_id))
+            time.sleep(1)
+            waited += 1
```
gregsfortytwo: Should probably use a start time and actually look at the clock instead of relying on an increment...it's unlikely to be a serious problem since this is just a timeout but they get inaccurate surprisingly quickly on many systems.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,782 @@
...
+        Return PID as an integer or None if not found
+        """
+        ps_txt = self.controller.run(
+            args=["ps", "ua", "-C", "ceph-fuse"],
+            check_status=False  # ps returns err if nothing running so ignore
+        ).stdout.getvalue().strip()
+        lines = ps_txt.split("\n")[1:]
+
+        for line in lines:
+            if line.find("--name client.{0} ".format(self.client_id)) != -1:
```
gregsfortytwo: Does the default client command match this string somehow, or is that a personal invocation convention?

jcsp: The client is always started by LocalFuseMount, so it's invoked consistently with --name

tasks/cephfs/vstart_runner.py
jcsp: In teuthology.orchestra there is a whole load of pipe handling that just doesn't exist here, where callers can pass in run.PIPE.  We can probably just assert that this is a string (it always is for the cases we care about in the cephfs tests)

tasks/cephfs/vstart_runner.py
jcsp: Does vstart actually do anything with pid files?  stop.sh is just a killall.

The spirit of this comment is that where we want services running for the duration of the test, it would be simpler to just have them as child processes of vstart_runner, so that we can do things like sending signals and then blocking on their actual termination, rather than polling to see things die etc.

tasks/cephfs/vstart_runner.py
jcsp: This is true, but there are a tonne of other places in the tests that we use this pattern, so I'm inclined to ignore until it becomes an issue.  One day a very bored person will convert all the timeouts to use the one helper function :-)

tasks/cephfs/vstart_runner.py
gregsfortytwo: Hmm, it's pretty dense but I thought stop.sh did a little more than that. In particular though you can also use init-ceph stop with a vstart cluster. I've switched back and forth over time; init-ceph used to be much better if you were sharing a box with somebody.

jcsp: oh, didn't know about init-ceph with a vstart cluster.  I think i'd defer doing nicer init handling until deciding either to go that way or to just start the processes directly from python-land.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,788 @@
...
+            args=[
+                'mkdir',
+                '--',
+                self.mountpoint,
+            ],
+        )
+
+        def list_connections():
+            self.client_remote.run(
+                args=["sudo", "mount", "-t", "fusectl", "/sys/fs/fuse/connections", "/sys/fs/fuse/connections"],
```
gregsfortytwo: Doesn't this sudo get stripped out?

jcsp: Indeed, harmlessly so.  removed.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,788 @@
...
+
+    @property
+    def _prefix(self):
+        return BIN_PREFIX
+
+    def set_clients_block(self, blocked, mds_id=None):
+        raise NotImplementedError()
+
+    def get_pgs_per_fs_pool(self):
+        # FIXME: assuming there are 3 OSDs
```
gregsfortytwo: Who uses 3 OSDs on a vstart cluster for the fs? 1, baby! :)
(Of course, this will still work for that, so...meh.)

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,788 @@
...
+    def set_ceph_conf(self, subsys, key, value):
+        self._conf[subsys][key] = value
+        self._write_conf()
+
+    def clear_ceph_conf(self, subsys, key):
+        del self._conf[subsys][key]
+        self._write_conf()
+
+    def clear_firewall(self):
+        # FIXME: unimplemented
```
gregsfortytwo: This should probably assert if it gets activated.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,788 @@
...
+    # Make sure the filesystem created in tests has uid/gid that will let us talk to
+    # it after mounting it (without having to  go root).  Set in 'global' not just 'mds'
+    # so that cephfs-data-scan will pick it up too.
+    filesystem.set_ceph_conf("global", "mds root ino uid", "%s" % os.getuid())
+    filesystem.set_ceph_conf("global", "mds root ino gid", "%s" % os.getgid())
+
+    # Monkeypatch get_package_version to avoid having to work out what kind of distro we're on
+    def _get_package_version(remote, pkg_name):
+        # Used in cephfs tests to find fuse version.  Your development workstation *does* have >=2.9, right?
+        return "2.9"
```
gregsfortytwo: ...eek. This needs to at least be something we can set via a config, please!

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,788 @@
...
+            is_named = False
+            for named in modules:
+                if named.endswith(method.id()):
+                    is_named = True
+                    break
+
+            if not is_named:
+                victims.append((case, method))
+
+    log.info("Disabling {0} tests because they're marked as long running".format(len(victims)))
```
gregsfortytwo: This output string is missing a case.

tasks/cephfs/vstart_runner.py
```diff
@@ -0,0 +1,788 @@
...
+        )
+
+    # Filter out tests that don't lend themselves to interactive running,
+    victims = []
+    for case, method in enumerate_methods(overall_suite):
+        fn = getattr(method, method._testMethodName)
+
+        drop_test = False
+
+        if hasattr(fn, 'is_long_running') and getattr(fn, 'is_long_running') is True:
```
gregsfortytwo: Can we have a switch to do the long-running tests on vstart as well, without explicitly naming them?

jcsp: Maybe in another patch?  Because this is aimed exclusively at developers I would expect people to write whatever changes they find they need.

tasks/cephfs/vstart_runner.py
jcsp: I'd sooner wait until someone actually needs to run on a super-old fuse (might never happen).  This is meant to just be a hack to get the filesystem tests running

tasks/cephfs/vstart_runner.py
jcsp: This is called during CephFSTestCase.tearDown, so pass is the desired behaviour.  If anyone tries to run a test that actually sets any firewall rules they'll fail out when trying to run iptables.

tasks/cephfs/vstart_runner.py
jcsp: changed.

tasks/cephfs/vstart_runner.py
gregsfortytwo: This is me being greedy because I'm not sure OS X Fuse supports all the 2.9
functions. :p

jcsp: Sounds like you're the right person to test and fix it!  I'm all on linux these days.

jcsp:
ping @gregsfortytwo -- no outstanding issues on this from my point of view

gregsfortytwo:
Yeah, it's in testing.

gregsfortytwo:
I'm seeing persistent issues with detecting the ceph-fuse admin socket in test runs which include this branch (and they aren't showing up elsewhere). See eg
/ubuntu-2015-09-09_07:03:51-fs-greg-fs-testing-testing-basic-multi runs 1047280,1047302,1047302. Each of them includes a client log of ceph-fuse running over the error window. There are some changes to the detection code but I don't see how it's breaking — on the other hand I don't really understand the glob code there anyway. :/

jcsp:
This now includes moving mds_journal_migration and mds_scrub_checks into CephFSTestCase subclasses, and the needed wait_for_daemons in the mds_journal_migration that was failing.

gregsfortytwo:
@jcsp 
http://pulpito.ceph.com/ubuntu-2015-09-29_00:30:50-fs-greg-fs-testing---basic-multi/1076047/

>2015-09-29T13:26:53.244 INFO:teuthology.orchestra.run.plana28.stderr:2015-09-29 13:26:53.232128 7f3d2ba62840  4 19273 events
>2015-09-29T13:26:53.302 INFO:tasks.cephfs.filesystem:Ran ['cephfs-journal-tool', '--debug-mds=4', '--debug-objecter=1', 'event', 'get', 'json', '--path', '/tmp/journal.json'] in time 0:00:04.812793, result:
>
>2015-09-29T13:26:53.303 INFO:teuthology.orchestra.run.burnupi42:Running: 'python -c \'import json; print len(json.load(open(\'"\'"\'/tmp/journal.json\'"\'"\')))\''
>2015-09-29T13:26:53.319 INFO:teuthology.orchestra.run.burnupi42.stderr:Traceback (most recent call last):
>2015-09-29T13:26:53.319 INFO:teuthology.orchestra.run.burnupi42.stderr:  File "<string>", line 1, in <module>
>2015-09-29T13:26:53.319 INFO:teuthology.orchestra.run.burnupi42.stderr:IOError: [Errno 2] No such file or directory: '/tmp/journal.json'
>2015-09-29T13:26:53.357 INFO:tasks.cephfs_test_runner:test_journal_migration (tasks.cephfs.test_journal_migration.TestJournalMigration) ... ERROR

jcsp:
Och, silly me.  Converting it to use the journal_tool wrapper means it wasn't running on the client machine any more (obviously works when I'm running it all on a single host).

gregsfortytwo:
I actually merged this to the infernalis branch and then merged that to master. Hurray!

PR #515, opened 2015-07-28
thrashosds: send a random osd a signal.SIGHUP

andrewschoen:
See: http://tracker.ceph.com/issues/12476

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -355,9 +355,9 @@ class Thrasher:
         if osd is None:
             osd = random.choice(self.dead_osds)
         self.log("Reviving osd %s" % (str(osd),))
-        self.live_osds.append(osd)
-        self.dead_osds.remove(osd)
         self.ceph_manager.revive_osd(osd, self.revive_timeout)
+        self.dead_osds.remove(osd)
+        self.live_osds.append(osd)
 
     def out_osd(self, osd=None):
         """
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -1748,6 +1748,13 @@ class CephManager:
         """
         self.raw_cluster_cmd('osd', 'in', str(osd))
 
+    def signal_osd(self, osd, sig):
+        """
+        Wrapper to local get_daemon call which sends the given
+        signal to the given osd.
+        """
+        self.ctx.daemons.get_daemon('osd', osd).signal(sig)
+
     ## monitors
     def signal_mon(self, mon, sig):
         """
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -5,6 +5,7 @@ from cStringIO import StringIO
 from functools import wraps
 import contextlib
 import random
+import signal
 import time
 import gevent
 import base64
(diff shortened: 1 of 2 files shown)


tasks/ceph_manager.py
```diff
@@ -135,6 +137,8 @@ def tmp(x):
             manager.raw_cluster_cmd('--', 'mon', 'tell', '*', 'injectargs',
                                     '--mon-osd-down-out-interval 0')
         self.thread = gevent.spawn(self.do_thrash)
+        if self.hangup:
```
zmc: ``self.hangup_delay``

tasks/ceph_manager.py
```diff
@@ -425,6 +429,8 @@ def do_join(self):
         """
         self.stopping = True
         self.thread.get()
+        if self.hangup:
```
zmc: ``self.hangup_delay``

dmick:
I'd recommend "hup" rather than "hangup" for familiarity.  looking deeper

tasks/ceph_manager.py
andrewschoen: ah, thanks!

tasks/ceph_manager.py
```diff
@@ -112,6 +113,7 @@ def __init__(self, manager, config, logger=None):
         self.clean_wait = self.config.get('clean_wait', 0)
         self.minin = self.config.get("min_in", 3)
         self.chance_move_pg = self.config.get('chance_move_pg', 1.0)
+        self.hangup_delay = self.config.get('hangup_delay')
```
zmc: might want to ``float()`` this

andrewschoen:
I'm fine with ``hup_delay`` over ``hangup_delay``. @zmc, thoughts?

tasks/ceph_manager.py
dmick: default to 0 explicitly?

andrewschoen: Good call, I won't here but when I use it in time.sleep I will.  I want a return value of None here to be ok.

zmc: I don't see an issue with defaulting to ``None``

tasks/thrashosds.py
```diff
@@ -112,6 +115,8 @@ def task(ctx, config):
         config = {}
     assert isinstance(config, dict), \
         'thrashosds task only accepts a dict for configuration'
+    # add default value for hangup_delay
+    config['hangup_delay'] = config.get('hangup_delay', 0.1)
```
dmick: hm, now I see this default.  Do we always want this behavior for thrash?

zmc: that seems to be what's being asked for

liewegas: yes

andrewschoen:
Alright, I've changed everything from 'hangup' to 'sighup'.

dmick:
:thumbsup:

andrewschoen:
This looks to be working as expected.

http://pulpito.ceph.com/sage-2015-07-28_19:05:15-rados-wip-12465-distro-basic-multi/

PR #517, opened 2015-07-31
Adds option for placing ceph daemons in cgroups

ivotron:
This closes http://tracker.ceph.com/issues/12424

As part of the work for http://tracker.ceph.com/issues/12379 we need a way to provide (relative) hardware determinism, which we address by using cgroups. This commit extends the ceph task so that cgroups can be created and mon/osd/mds daemons can be placed in them.

I've tested this in the cephlab on mira machines (with cgroups enabled and disabled) 

Related PR: https://github.com/ceph/ceph-cm-ansible/pull/94

Note that this will work only for ubuntu systems unless the changes in https://github.com/ceph/ceph-cm-ansible/pull/94 get ported to other distros.

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -760,6 +760,146 @@ def cluster(ctx, config):
             ),
         )
 
+    if config['cgroups']:
+        log.info('Creating cgroups...')
+        def has_net_limits(remote, roles, config):
+            # checks if a cgroup configuration for a role in the given remote
+            # contains the 'net.limit' parameter
+            for role in roles:
+                if role in config['cgroups'] and config['cgroups'][role]:
+                    for param in config['cgroups'][role]:
+                        if 'net.limit' in param:
+                            return True
+            return False
+        for remote, roles_for_host in ctx.cluster.remotes.iteritems():
+            remote.run(
+                args=[
+                    'sudo',
+                    'cgcreate',
+                    '-g',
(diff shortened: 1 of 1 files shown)


ivotron:
An alternative to the changes in https://github.com/ceph/ceph-cm-ansible/pull/94 (that got reverted by https://github.com/ceph/ceph-cm-ansible/pull/95) is to add the packages explicitly in an ansible task, for example:

```yaml
roles:
- [mon.0, osd.0, osd.1, osd.2, client.0]
tasks:
- ansible:
    repo: https://github.com/ceph/ceph-cm-ansible.git
    playbook:
    - tasks:
      - name: "Install cgroup-lite"
        apt: name=cgroup-lite state=present
      - name: "Install cgroup-bin"
        apt: name=cgroup-bin state=present
      - name: "load net_cls cgroups module"
        modprobe: name=cls_cgroup state=present
      - name: "re-mount cgroups virtual filesystem"
        service: name=cgroup-lite state=restarted
    hosts:
    - client.0
- install:
- ceph:
    cgroups:
      osd.0: [io.limit=40000000]
- radosbench:
    clients: [client.0]
    time: 60
```

should this be part of the documentation?

liewegas:
I think teh plan is to have systemd starting/stopping daemons instead of ceph.py's snowflake way to running the cluster.  In that case we'll want to be twiddling the systemd cgroup-related options, I assume?

Closing this for now..

PR #518, opened 2015-08-03
Fix experimental feature override makes user config useless

yuyuyu101:
Signed-off-by: Haomai Wang haomaiwang@gmail.com

### DIFF
--- a/erasure-code/ec-rados-plugin=shec-k=4-m=3-c=2.yaml
+++ b/erasure-code/ec-rados-plugin=shec-k=4-m=3-c=2.yaml
@@ -26,5 +26,4 @@ overrides:
   ceph: 
     conf: 
       global: 
-        enable experimental unrecoverable data corrupting features: shec 
-
+        enable experimental unrecoverable data corrupting features: shec, ms-type-async


dachary:
That will make shec tests fail.

yuyuyu101:
@dachary do you have other ideas? Otherwise, I can't run async-messenger tests which reference to this yaml.

athanatos:
Annoying.  You modify the ceph task to check for an experiemental_features list at the same level as the config dict.  Then, the override merge logic would give you the union of the lists and would presumably behave better.  You could probably also just add async messenger to everywhere that mentions the expermental features config value.

yuyuyu101:
@athanatos updated

dachary:
@yuyuyu101 looks great :-) Could you run at least one job including this change and append the URL of the result here ? When done it's good to merge.

yuyuyu101:
@dachary it(http://pulpito.ceph.com/haomai-2015-08-04_20:31:08-rados-master-distro-basic-multi/) showed no failed job related to this job. :-)

dachary:
@yuyuyu101 could you please link to the specific job that contains the file modified by this patch ?

yuyuyu101:
@dachary just like this(http://pulpito.ceph.com/haomai-2015-08-04_20:31:08-rados-master-distro-basic-multi/999879/) it will have "shec ms-type-async"

dachary:
@yuyuyu101 is there a successful job with **shec ms-type-async** ?

liewegas:
On Wed, 5 Aug 2015, Haomai Wang wrote:
> @dachary just like this(http://pulpito.ceph.com/haomai-2015-08-04_20:31:08-rados-master-distro-basic-multi/999879/) it will have "shec
> ms-type-async"

we could make a (perhaps undocumented?) feature that a '*' value will 
match any experimental feature for the purposes of qa...

yuyuyu101:
@dachary like this(http://pulpito.ceph.com/haomai-2015-08-04_20:31:08-rados-master-distro-basic-multi/999905/) :-)

PR #520, opened 2015-08-05
Set the SELinux context of the logrotate config

zmc:
This is for http://tracker.ceph.com/issues/12621


ktdreyer:
Why are we using a custom log rotation file instead of using the one in the packaging?

ktdreyer:
(This strikes me as the sort of pattern where Teuthology does some special workflow that ends up masking bugs that real users would encounter)

PR #521, opened 2015-08-06
tasks: fix syntax error in ceph.py

jcsp:
From e195f9fa.


tchaikov:
`Reviewed-by: Kefu Chai <kchai@redhat.com>`

PR #526, opened 2015-08-10
tasks/cephfs: implement TestDataScan.test_stashed_layout

jcsp:
This is for testing how cephfs-data-scan handles
the new 'layout' xattr when found during recovery.


jcsp:
This tests (and depends on) the code in https://github.com/ceph/ceph/pull/5523

PR #529, opened 2015-08-12
calamari_setup: install "cli" utils on Calamari node

ktdreyer:
In our RHCS 1.3 ceph-deploy docs, we tell users to run "ceph-deploy install --cli" on their calamari admin node, but our smoke test wasn't actually doing this.

See https://bugzilla.redhat.com/1252929 , "[Ubuntu 1.3.0] - ceph-deploy install --no-adjust-repos --cli `hostname` is failing with a Traceback error"

### DIFF
--- a/tasks/calamari_setup.py
+++ b/tasks/calamari_setup.py
@@ -377,6 +377,8 @@ def deploy_ceph(ctx, cal_svr):
                     ' '.join(all_osds))
         cmds.append('ceph-deploy install --no-adjust-repos --osd ' +
                     ' '.join(all_osds))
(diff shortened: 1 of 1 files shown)


ktdreyer:
cc'ing @trhoden for review

trhoden:
Seems reasonable so long as the teuthology nodes can handle the call to \`hostname\`.

ktdreyer:
I've run a calamari_setup test with this branch, and it does the right thing.

ktdreyer:
I've amended my original commit (6948df52546714a996fd0f48db2e61fb2c0e9775) to include a comment explaining that `hostname` comes directly from our docs

PR #537, opened 2015-08-18
update the list of supported distributions

dachary:
CentOS 6 and Ubuntu 12.04 are not support after Infernalis.
CentOS 7 is supported starting with Infernalis.


dachary:
I ssupect we want to wait until http://tracker.ceph.com/issues/12725 is fixed before merging this.

PR #541, opened 2015-08-26
fix rebug rados

XinzeChi:
Signed-off-by: Xinze Chi xmdxcxz@gmail.com


XinzeChi:
@dachary , Do you think this is spelling mistake?

dachary:
@XinzeChi nice catch !

PR #543, opened 2015-08-27
suites/rados/thrash: reduce rados bench run time

zhouyuan:
1800 sec was a bit too long and also generates too many data. Using
a short run would make the rados suite faster.
Note the upgrade suite still uses the long run(1800 sec) to produce
enough stress.

Signed-off-by: Yuan Zhou <yuan.zhou@intel.com>


gregsfortytwo:
@athanatos do we need the 30-minute run times here? I thought they were set that long because that's how long it takes for some of the bugs we discover, but maybe not.

liewegas:
@athanatos ping

dachary:
@athanatos ping ?

PR #549, opened 2015-08-28
Add rbd fio test to exercise block IO on all format/features

vasukulkarni:
This test can be used to exercise all rbd supported formats/features and can be used along with thrash/powercuts,  Can be used with large disks and simulate VM installation test cases. I am still testing this with other thrash tests for complete results.

### DIFF
--- a/tasks/rbd_fio.py
+++ b/tasks/rbd_fio.py
@@ -1,0 +1,201 @@
+"""
+ Long running fio tests on rbd mapped devices for format/features provided in config
+ Many fio parameters can be configured so that this task can be used along with thrash/power-cut tests
+ and exercise IO on full disk for all format/features
+  - This test should not be run on VM due to heavy use of resource
+  
+"""
+import contextlib
+import logging
+import StringIO
+import re
+
+from teuthology.parallel import parallel
+from teuthology import misc as teuthology
+from tempfile import NamedTemporaryFile
+from teuthology.orchestra import run
+
+log = logging.getLogger(__name__)
+
+@contextlib.contextmanager
+def task(ctx, config):
+    """
+    client.0:
+       fio-io-size: 100g or 80% or 100m
+       fio-version: 2.2.9
+       formats: [2]
+       features: [1, 2, 13]
+       test-clone-io: 1  #remove this option to not run create rbd clone and not run io on clone
+       io-engine: "sync or rbd or any io-engine"
+       rw: randrw
+    client.1:
+       fio-io-size: 100g
+       fio-version: 2.2.9
+       rw: read
+       image-size:20480 
+
+or
+    all:
+       fio-io-size: 400g
+       rw: randrw
+       formats: [2]
+       features: [1,2]
+       io-engine: libaio
+       
+    Create rbd image + device and exercise IO for format/features provided in config file
+    Config can be per client or one config can be used for all clients, fio jobs are run in parallel for client provided
+    
+    """
+    if config.get('all'):
+        client_config = config['all']  
+    clients = ctx.cluster.only(teuthology.is_type('client'))
+    for remote,role in clients.remotes.iteritems():
+        if 'client_config' in locals():
+           with parallel() as p: 
+               p.spawn(run_fio, remote, client_config)
+        else:    
+           for client_config in config:
+              if client_config in role:
+                 with parallel() as p:
+                     p.spawn(run_fio, remote, config[client_config])
+            
+    yield
+    
+    
+def run_fio(remote, config):
+    """
+    create fio config file with options based on above config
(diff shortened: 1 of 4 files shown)


vasukulkarni:
@jdurgin @dillaman: can you review and let me know what format/features to add for config
1st yaml run log:  http://paste.ubuntu.com/12240259/
2nd task with thrash: http://paste.ubuntu.com/12240236/

vasukulkarni:
@dillaman : fio doesn't have a good binaries for all distro's so thats the reason I am using the source, for the librbd enginet, I think I only need librbd-devel to be installed for it to work and probably has to be compiled from source,  I will try to get a good run with librbd engine and update here.

vasukulkarni:
@dillaman @jdurgin: Added test for librbd using rbd engine (format 2, features: 1,2,4) 
Logs: http://paste.ubuntu.com/12267641/

vasukulkarni:
@jdurgin @dillaman I am planning to merge this after I get a good run, Let me know if you need any more changes, it now tests on librbd clones using test-clone-io option, right now there are failures which seem unrelated to scripts here http://pulpito.ceph.com/ubuntu-2015-09-30_17:21:12-rbd-infernalis---basic-multi/ and I have also scheduled a krbd run.

tasks/rbd_fio.py
```diff
@@ -0,0 +1,202 @@
...
+           remote.run(args=['sudo', 'rbd', 'info', 'i{i}f{f}{sn}'.format(i=frmt,f=feature,sn=sn)])
+           if ioengine != 'rbd':
+               out=StringIO.StringIO()
+               remote.run(args=['sudo', 'rbd', 'map', 'i{i}f{f}{sn}'.format(i=frmt,f=feature,sn=sn)],stdout=out)
+               dev=re.search(r'(/dev/rbd\d+)',out.getvalue())
+               rbd_dev=dev.group(1)
+               if config.get('test-clone-io'):
+                    log.info("Testing clones using fio")
+                    remote.run(args=['sudo', 'rbd', 'snap', 'create', 'i{i}f{f}{sn}@i{i}f{f}{sn}Snap'.format(i=frmt,f=feature,sn=sn)])
+                    remote.run(args=['sudo', 'rbd', 'snap', 'protect', 'i{i}f{f}{sn}@i{i}f{f}{sn}Snap'.format(i=frmt,f=feature,sn=sn)])
```
zmc: I notice repeating the same string templates many times. Wouldn't it be nicer to only do that once?

tasks/rbd_fio.py
```diff
@@ -0,0 +1,202 @@
...
+
+from teuthology.parallel import parallel
+from teuthology import misc as teuthology
+from tempfile import NamedTemporaryFile
+from teuthology.orchestra import run
+
+log = logging.getLogger(__name__)
+
+@contextlib.contextmanager
+def task(ctx, config):
```
zmc: Curious, why not use ``teuthology.task.Task`` ?

tasks/rbd_fio.py
vasukulkarni: yes agree with the repetition, will fix that.

tasks/rbd_fio.py
```diff
@@ -0,0 +1,202 @@
...
+                     p.spawn(run_fio, remote, config[client_config])
+            
+    yield
+    
+    
+def run_fio(remote, config):
+    """
+    create fio config file with options based on above config
+    get the fio from github, generate binary, and use it to run on
+    the generated fio config file
```
zmc: If it were me, for easier readability and maintenance, I would want to split this function up into the logical sections outline in this docstring.

tasks/rbd_fio.py
vasukulkarni: yeah sorry, I always wanted to use the new Task style but just slipped, I will start using the new Task style going forward.

zmc:
hmm. merging your own PR would seem to defeat the purpose of review...

vasukulkarni:
@zmc feel free to undo this and let me know if you have more comments to fix, I had some discussion with @jdurgin before and I have been testing this for quite some time.

zmc:
@vasukulkarni I don't see any other feedback here aside from mine and I didn't consider it resolved.

If everyone else is happy with it, fine. Unlike in ``teuthology``, I don't really end up maintaining this code.

vasukulkarni:
yeah sorry my bad, I think I should have mentioned @jdurgin is OK with this test and I added clone on io based on his feedback.

jdurgin:
for the record I said I was ok with the rbd coverage prior to zack's review, I didn't look at the task itself too closely, and didn't expect this to be merged without resolving zack's comments

PR #552, opened 2015-09-03
tasks/scrub_test: sudo ls ...

liewegas:
/var/lib/ceph/osd/* is owned by ceph now; ubuntu user can't read
it.

Fixes: #12878


tchaikov:
testing at http://pulpito.ceph.com/kchai-2015-09-03_02:54:08-rados-master---basic-multi/1043139/

tchaikov:
lgtm and test passed.

PR #555, opened 2015-09-06
rados/thrash-erasure-code-isa: use supported distros

liewegas:
Was trusty and rhel7; instead use supported list (currently
centos7 and trusty, will soon include jessie and maybe fedora).

Signed-off-by: Sage Weil sage@redhat.com


liewegas:
@dachary I'm not sure if the distro choice used before was that way for a specific reason?

dachary:
The important bit is **arch: x86_64** which can be moved to an arch/arch.yaml file instead to apply to each distro instead. More flexible :-)

PR #556, opened 2015-09-08
tasks/ceph_manager: dump pgs when recover times out

liewegas:
It is really hard to map a stuck recovery back to the pgs that
are stuck.  This will make it easy.

Signed-off-by: Sage Weil sage@redhat.com


dachary:
Excellent idea. I would go as far as running **ceph report** even but that's a bit much ;-)
<pre>
</pre>

PR #558, opened 2015-09-10
rados: 'failed to encode ...' warnings are normal on upgrades

liewegas:
Fixes: #12973


tchaikov:
tested at http://pulpito.ceph.com/kchai-2015-09-10_00:06:01-rados-master---basic-multi/

PR #559, opened 2015-09-11
leaks

liewegas:


liewegas:
@zmc look ok?

tasks/ceph.py
```diff
@@ -246,8 +246,12 @@ def valgrind_post(ctx, config):
                 log.error('saw valgrind issue %s in %s', kind, file)
                 valgrind_exception = Exception('saw valgrind issues')
 
-        if valgrind_exception is not None:
-            raise valgrind_exception
+        if config.get('expect-valgrind-errors'):
```
zmc: Is ``expect-valgrind-errors`` a new setting? If so, I'd suggest using underscores if you don't mind.

liewegas: On Fri, 11 Sep 2015, Zack Cerza wrote:
> 
> In tasks/ceph.py:
> 
> > @@ -246,8 +246,12 @@ def valgrind_post(ctx, config):
> >                  log.error('saw valgrind issue %s in %s', kind, file)
> >                  valgrind_exception = Exception('saw valgrind issues')
> >  
> > -        if valgrind_exception is not None:
> > -            raise valgrind_exception
> > +        if config.get('expect-valgrind-errors'):
> 
> Is expect-valgrind-errors a new setting? If so, I'd suggest using underscores if you don't mind.

fixed!

PR #563, opened 2015-09-11
fix test for 11429 (and now 13060)

liewegas:
This verifies the fix for 13060 and also the similar fix for infernalis.


liewegas:
passes w/ updated fix

PR #565, opened 2015-09-14
buildpackages: build ceph packages and upload them

dachary:
Build Ceph packages from source for the required revision, os_type,
os_version and architecture and upload them to the gitbuilder
repository.

http://tracker.ceph.com/issues/13031 Fixes: #13031

### DIFF
--- a/suites/teuthology/buildpackages/buildpackages.yaml
+++ b/suites/teuthology/buildpackages/buildpackages.yaml
@@ -1,0 +1,5 @@
+roles:
+    - [mon.0, client.0]
+tasks:
+    - buildpackages:
+    - install:
--- a/suites/teuthology/buildpackages/ubuntu_14.04.yaml
+++ b/suites/teuthology/buildpackages/ubuntu_14.04.yaml
@@ -1,0 +1,1 @@
+../../../distros/supported/ubuntu_14.04.yaml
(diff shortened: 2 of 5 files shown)


dachary:
@joejulian FYI, the stub doing nothing right now

gregsfortytwo:
<without checking the code in any detail>

Does this mean somebody including this task on the sepia lab would be building packages on the main teuthology box? (I'm pretty sure.)
...or, no, it looks like it only works for openstack clusters? And why do it on a per-job basis instead of per-suite-run?

I'm a big fan of not having to wait for the gitbuilders or whatever, but I think this approach is a very expensive one (packages take like 30 minutes on a dedicated box, right?).

dachary:
@gregsfortytwo this would build packages exactly once for a given sha1 / os_type / os_version / architecture. It is triggered on a per job basis but the results are shared between all jobs in the cluster.

dachary:
> it looks like it only works for openstack clusters?

Yes. And it would work in the sepia lab as well provided the sepia lab has access to an OpenStack cluster.

dachary:
> packages take like 30 minutes on a dedicated box, right?

Yes. It can be optimized down to 15 minutes but that's not what this pull request does.

PR #566, opened 2015-09-16
Use old branches as is without 'ceph-' for firefly, hammer etc

vasukulkarni:
Use old branches as is without 'ceph-' for firefly, hammer etc


gregsfortytwo:
Looks like the test failures are only in buildpackages.py, which is not the fault of this PR.

PR #567, opened 2015-09-16
ceph: wait for CephFS to be healthy before proceeding

gregsfortytwo:
Ceph PR https://github.com/ceph/ceph/pull/5416 was causing tests to occasionally fail because it took some time for the MDSes to actually be ready to go. Wait for them to be happy during initial setup.

We set a 300 second timeout just to make sure we don't eat up machines for forever.

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -292,6 +292,9 @@ def cephfs_setup(ctx, config):
             'ceph',
             'mds', 'set_max_mds', str(num_active)])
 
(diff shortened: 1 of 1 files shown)


gregsfortytwo:
There was an accidental commit of code which doesn't pass pyflakes, but this PR is clean in its own changes.

ukernel:
This RP causes http://tracker.ceph.com/issues/13129

PR #570, opened 2015-09-17
tasks/ceph: fix waiting for MDS active

jcsp:
Do it in healthy() which runs after daemons
start.  Fixes 15fe7acb.


gregsfortytwo:
I've just reverted the waiting merge for now, will pull this in through other means.

gregsfortytwo:
Pushed to master as eab70197a67397797ce7bd81c667900dbecdabd4 with a reworded commit message.

PR #573, opened 2015-09-21
hammer-x fixes

liewegas:


suites/upgrade/hammer-x/stress-split-erasure-code-x86_64/distros/ubuntu_14.04.yaml
```diff
@@ -1,3 +0,0 @@
-os_type: ubuntu
-os_version: "14.04"
-arch: x86_64
```
dachary: You need an arch directory like in https://github.com/ceph/ceph-qa-suite/pull/571/files

dachary:
makes sense to me :-) 
<pre>
</pre>
with the arch directory and once a green run confirms it actually works

suites/upgrade/hammer-x/stress-split-erasure-code-x86_64/distros/ubuntu_14.04.yaml
liewegas: merged the other one

PR #577, opened 2015-09-24
tolerate pg_num get failure

liewegas:
This races with pool deletion sometimes.

I know I should catch the exeption by name but I'm not sure how to tell
what it is named...

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -19,6 +19,7 @@ from tasks.scrub import Scrubber
 from util.rados import cmd_erasure_code_profile
 from teuthology.orchestra.remote import Remote
 from teuthology.orchestra import run
+from teuthology.exceptions import CommandFailedError
 
 
 DEFAULT_CONF_PATH = '/etc/ceph/ceph.conf'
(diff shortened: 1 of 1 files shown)


tasks/ceph_manager.py
```diff
@@ -793,7 +793,11 @@ def tmp(x):
         pools = self.list_pools()
         self.pools = {}
         for pool in pools:
-            self.pools[pool] = self.get_pool_property(pool, 'pg_num')
+            # we may race with a pool deletion; ignore failures here
+            try:
+                self.pools[pool] = self.get_pool_property(pool, 'pg_num')
+            except:
```
zmc: This should be as specific as possible; bare ``except:`` clauses should never be necessary. Here's why the're bad:
https://docs.python.org/2/howto/doanddont.html#except

tasks/ceph_manager.py
```diff
@@ -793,7 +793,11 @@ def tmp(x):
...
+                pass
```
zmc: I'd recommend at least logging something here, or else you'll never know when this is happening. Unless you really don't want to know :)

tasks/ceph_manager.py
```diff
@@ -793,7 +793,12 @@ def tmp(x):
         pools = self.list_pools()
         self.pools = {}
         for pool in pools:
-            self.pools[pool] = self.get_pool_property(pool, 'pg_num')
+            # we may race with a pool deletion; ignore failures here
+            try:
+                self.pools[pool] = self.get_pool_property(pool, 'pg_num')
+            except Exception:
+                self.log.info('Failed to get pg_num rom pool %d, ignoring' % pool)
+                pass
```
zmc: Don't need the ``pass`` anymore

PR #579, opened 2015-09-25
tasks/cephfs: fix cap cleardown between tests

jcsp:
This used to remove the mds cap.  Now that cap validation
is working better, the cluster won't tolerate this.


jcsp:
Tested on a vstart cluster.

PR #580, opened 2015-09-26
ceph-deploy: fix ceph-deploy-branch config options

dachary:
The config paramter of download_ceph_deploy does not have a ceph-deploy
item, therefore the ceph-deploy-branch parameter is always assumed to be
master.


dachary:
@jdurgin would you have time to look at this simple fix ?

dachary:
@jdurgin thanks !

PR #581, opened 2015-09-26
buildpackages: get Ceph git url via get_ceph_git_url()

dachary:
Instead of relying on git_base_url, use the get_ceph_git_url() to obtain
the URL of the Ceph repository to use with git clone. This allows the
user to override it via the git_ceph_url configuration option and the
--git-ceph-url command line option to teuthology-openstack.

http://tracker.ceph.com/issues/11883 Refs: #11883


dachary:
*pass* http://167.114.234.211:8081/ubuntu-2015-09-26_11:15:29-ceph-disk-wip-12936-shec---basic-openstack/34/

PR #583, opened 2015-09-26
buildpackages: flock must --close

dachary:
The process run by flock must not inherit the file descriptor because
this will cause the lock to be held forever should the command survive
the call to flock. This is for instance the case for the ssh-agent.


dachary:
**pass** http://167.114.226.190:8081/ubuntu-2015-09-26_13:31:37-teuthology:buildpackages-master---basic-openstack/

PR #584, opened 2015-09-27
buildpackages: user-data must be per os-type/os-version

dachary:
CentOS 6.5 needs to install a package and reboot to grow the root file
system. Instead of assuming a common user-data.txt file can fit all
Operating Systems, make one user data per os-type/os-version combination.


dachary:
**pass** http://167.114.226.190:8081/ubuntu-2015-09-27_21:57:48-teuthology:buildpackages-hammer---basic-openstack/

PR #586, opened 2015-09-28
tasks/cephfs: mds allow

liewegas:


jcsp:
This was also https://github.com/ceph/ceph-qa-suite/pull/579 and https://github.com/ceph/ceph-qa-suite/pull/587, closing those.

PR #589, opened 2015-09-29
s3tests using ec data pool for smoke

vasukulkarni:
Add s3tests using ec data pool


vasukulkarni:
@yehudasa we can add this to smoke and will get tested early enough during wip since rgw is a bigger suite.

vasukulkarni:
@yehudasa : do you think it makes sense to add this to smoke? s3 test with ec data pool and civetweb.

yehudasa:
@vasukulkarni yeah, we can put it in the smoke test

vasukulkarni:
@yehudasa thanks, I have tested this.

PR #591, opened 2015-09-30
rados: add test for 13234.yaml

athanatos:


athanatos:
DNM yet, haven't tested this version of the test.

athanatos:
Ok, verified that it works, good to merge once reviewed.

PR #606, opened 2015-10-05
tasks/ceph_fuse.py: virtual machines need flexible mount timeout

dachary:
can give them 60 seconds.
for VM tests.

http://tracker.ceph.com/issues/12320 Fixes: #12320

(cherry picked from commit ec12f21a7ead9d82b575c4f927d5e7b73f85c3d5)

### DIFF
--- a/tasks/cephfs/fuse_mount.py
+++ b/tasks/cephfs/fuse_mount.py
@@ -100,11 +100,16 @@ class FuseMount(CephFSMount):
         self.fuse_daemon = proc
 
         # Wait for the connection reference to appear in /sys
+        mount_wait = self.client_config.get('mount_wait', 0)
+        if mount_wait > 0:
+            log.info("Fuse mount waits {0} seconds before checking /sys/".format(mount_wait))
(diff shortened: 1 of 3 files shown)


tasks/cephfs/fuse_mount.py
```diff
@@ -100,11 +100,16 @@ def list_connections():
         self.fuse_daemon = proc
 
         # Wait for the connection reference to appear in /sys
+        mount_wait = self.client_config.get('mount_wait', 0)
+        if mount_wait > 0:
+            log.info("Fuse mount waits {0} seconds before checking /sys/".format(mount_wait))
+            time.sleep(mount_wait)            
+        timeout = self.client_config.get('mount_timeout', '30')
```
gregsfortytwo: I'm not sure how the git history worked out here but see 8e275049a0a8b976fc9eb5c63dcefe01bedcc711

dachary:
@gregsfortytwo thanks for your review, the additional commit was cherry-picked and pushed :-)

PR #615, opened 2015-10-10
revert "cephfs/test_client_limits.py: update test_client_oldest_tid"

ukernel:
infernalis branch of ceph does not include the cap flush changes


gregsfortytwo:
This assessment was inaccurate, it contains the cap flush changes and we need to test them! :)

PR #619, opened 2015-10-12
buildpackages: get ceph submodules

dachary:
Prior to v0.80.9, autogen.sh did not get submodules. Copy/paste the
submodule initialization from newer autogen.sh in common.sh so that
v0.80.8 and below can be rebuilt from sources. It does not hurt to
update the submodules twice.


dachary:
http://tracker.ceph.com/issues/12386#note-23 successfully ran with it

PR #623, opened 2015-10-13
rados/singleton-nomsgr/all/msgr: needs 15GB RAM

dachary:
If given only 8GB RAM,  ceph_test_msgr may abort with buffer::bad_alloc.

http://tracker.ceph.com/issues/11260 Fixes: #11260


dachary:
The problem showed at http://tracker.ceph.com/issues/12386#note-22 with **rados/singleton-nomsgr/{all/msgr.yaml}** and with this patch (although applied on the hammer branch) it passed the same test as part of the full rados run at http://tracker.ceph.com/issues/12386#note-23

liewegas:
lgtm

PR #624, opened 2015-10-13
Wip 13437

ukernel:


jcsp:
The fix for #13437 looks good to me.  Is the "test ceph-fuse with libcephfs permission check" commit in here by accident?

PR #626, opened 2015-10-13
openstack: rados/thrash: allocate three disks, always

dachary:
The thrasher needs disk attached to run against xfs, ext4 or btrfs. And
some jobs use more disks and do not fit in 40GB.

http://tracker.ceph.com/issues/13450 Fixes: #13450

### DIFF
--- a/suites/rados/thrash/clusters/openstack.yaml
+++ b/suites/rados/thrash/clusters/openstack.yaml
@@ -1,0 +1,8 @@
+openstack:
+  machine:
+    disk: 40 # GB
+    ram: 8000 # MB
+    cpus: 1
+  volumes: # attached to each instance
+    count: 3
+    size: 30 # GB


dachary:
http://tracker.ceph.com/issues/12386#note-23 ran with this patch and 20GB attached disks and was mostly successfull with the exception of **rados/thrash/{0-size-min-size-overrides/2-size-2-min-size.yaml 1-pg-log-overrides/short_pg_log.yaml clusters/{fixed-2.yaml openstack.yaml} fs/ext4.yaml msgr-failures/fastclose.yaml thrashers/default.yaml workloads/radosbench.yaml}** which apparently requires more.

dachary:
Running again with 20GB attached disk at http://167.114.239.68:8081/ubuntu-2015-10-13_08:17:20-rados-hammer-backports---basic-openstack and with 30GB attached disks at (XXXX).

dachary:
@tchaikov the tests above pass. There seems to be a significant variance regarding how much disk space a given job is going to use. The first run with 20GB disks had one false negative because there was not enough disk. But re-running it passed ( http://167.114.239.68:8081/ubuntu-2015-10-13_08:17:20-rados-hammer-backports---basic-openstack ). Increasing to 30GB may help but we will probably only know if there are no false negative in the next 20 runs or so.

dachary:
The following runs used this pull request and passed (environmental noise re-run + 1 know bug that is unrelated) 
* http://167.114.239.68:8081/ubuntu-2015-10-13_17:19:15-rados-hammer-backports---basic-openstack
* http://167.114.239.68:8081/ubuntu-2015-10-14_06:38:42-rados-hammer-backports---basic-openstack

PR #630, opened 2015-10-13
upgrade/stress-split: needs only six OSDs

dachary:
The large number of OSDs is only needed for LRC or SHEC in some
cases. All other jobs will do fine with six OSDs.

http://tracker.ceph.com/issues/13408 Fixes: #13408


liewegas:
cool, lgtm (i don't see shec/lrc in the stress-split workloads)

dachary:
> i don't see shec/lrc in the stress-split workloads

An LRC test should be added but ... that's another issue.

PR #632, opened 2015-10-13
use cephtest as base dir during testing 

vasukulkarni:
http://tracker.ceph.com/issues/13431
use cephtest as base dir during testing and in case of failures let nuke cleanup any stale dir's

### DIFF
--- a/tasks/rbd_fio.py
+++ b/tasks/rbd_fio.py
@@ -49,20 +49,21 @@ or
     if config.get('all'):
         client_config = config['all']  
     clients = ctx.cluster.only(teuthology.is_type('client'))
+    rbd_test_dir = teuthology.get_testdir(ctx) + "/rbd_fio_test"
     for remote,role in clients.remotes.iteritems():
         if 'client_config' in locals():
            with parallel() as p: 
-               p.spawn(run_fio, remote, client_config)
+               p.spawn(run_fio, remote, client_config, rbd_test_dir)
         else:    
            for client_config in config:
               if client_config in role:
                  with parallel() as p:
-                     p.spawn(run_fio, remote, config[client_config])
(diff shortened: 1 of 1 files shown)


vasukulkarni:
@zmc @dmick : can one of you review and merge this.

tasks/rbd_fio.py
```diff
@@ -183,16 +183,17 @@ def run_fio(remote, config):
     try:
         log.info("Running rbd feature - fio test on {sn}".format(sn=sn))
         fio = "https://github.com/axboe/fio/archive/fio-" + fio_version + ".tar.gz"
-        remote.run(args=['mkdir', run.Raw('~/rbd-fio-test'),])
-        remote.run(args=['cd' , run.Raw('~/rbd-fio-test'),
+        rbd_test_dir = "~/cephtest/rbd-fio-test"  #let nuke cleanup in case of abrupt failures
```
zmc: Instead of hardcoding this, I would use this function:
https://github.com/ceph/teuthology/blob/master/teuthology/misc.py#L132

Just to avoid potential weirdness if that changes someday.

vasukulkarni: @zmc yep makes sense

tasks/rbd_fio.py
```diff
@@ -49,20 +49,21 @@ def task(ctx, config):
     if config.get('all'):
         client_config = config['all']  
     clients = ctx.cluster.only(teuthology.is_type('client'))
+    rbd_test_dir = teuthology.get_testdir(ctx) + "/rbd_fio_test"
```
dmick: (and it really ought to be os.path.join(teuthology.get_testdir(ctx), '/rbd_fio_test'))

PR #633, opened 2015-10-14
tasks/cephfs/test_failover

jcsp:
Test to go with https://github.com/ceph/ceph/pull/6253


tasks/cephfs/filesystem.py
```diff
@@ -220,6 +219,15 @@ def get_active_names(self):
```
gregsfortytwo: You need to actually search for the state string instead of 'up:active'. ;)

jcsp: Hmm?  Were you looking at are_daemons_healthy?  The change is to the get_daemon_names function where it is now comparing with the passed in state.

gregsfortytwo: I added a patch fixing it, you weren't using the passed-in value.

PR #646, opened 2015-10-19
ceph-deploy: use ceph-deploy repo instead of --dev*

dachary:


vasukulkarni:
@dachary shouldn't we keep the dev option, that is the documented version and I am running some workunits using that option and the dev options supports released builds and sha, Is the ceph-deploy cli also going to change?

dachary:
@vasukulkarni  instead of using
<pre>
ceph-deploy:
   branch:
      dev: mybranch
</pre>
you can use
<pre>
ceph-deploy:
   branch: mybranch
</pre>
which is consistent with the **install** task. Which workunit are using the former ? I searched ceph-qa-suite but did not find any.
I don't know what the plans are for **ceph-deploy**: I don't think there is a plan to deprecate --dev and --dev-commit.

vasukulkarni:
@dachary those tests are still in my private directory and in testing, I see the new option aligns with what we use for ceph task and repo option takes care of the setting the repo, I think you also need change in suite.py for ceph-deploy placeholder which still uses dev option.

dachary:
@vasukulkarni I have a commit for that change at https://github.com/ceph/teuthology/commit/b3925acf38d2a821d2a266bbc36054cdacc17a4b and everything needs to be synchronized which is kind of tricky :-)

dachary:
@vasukulkarni I've tested the combo ( https://github.com/ceph/teuthology/pull/661 + https://github.com/ceph/ceph-qa-suite/pull/646 + https://github.com/ceph/teuthology/pull/661 ) and as soon as @alfredodeza merges we can move on.

vasukulkarni:
@dachary sure, this LGTM, will wait for the ceph-deploy PR.

PR #648, opened 2015-10-19
test offline upgrade from firefly -> x

liewegas:


dzafman:
Nothing jumps out at me, so LGTM

yuriw:
@liewegas I thought we said it's an invalid upgrade path. Are we trying to allow f-x/infernalis now?

dzafman:
@yuriw @liewegas Yes, we are allowing Firefly to Infernalis but you have to restart the OSDs in that case.

PR #650, opened 2015-10-20
Added openstack.yaml config

yuriw:
@jcsp @dachary 

Testing now as

<pre>
teuthology-suite -v -c hammer -m openstack -s fs -k distro -p 70 --suite-branch wip-openstack-fixes-fs_hammer -e yweinste@redhat.com --filter="fs/basic/{clusters/{fixed-3-cephfs.yaml openstack.yaml} debug/mds_client.yaml fs/btrfs.yaml inline/no.yaml overrides/whitelist_wrongly_marked_down.yaml tasks/cfuse_workunit_kernel_untar_build.yaml}"
</pre>

http://pulpito.ovh.sepia.ceph.com:8081/teuthology-2015-10-20_18:16:50-fs-hammer-distro-basic-openstack/

### DIFF
--- a/suites/fs/basic/tasks/cfuse_workunit_kernel_untar_build.yaml
+++ b/suites/fs/basic/tasks/cfuse_workunit_kernel_untar_build.yaml
@@ -4,3 +4,11 @@ tasks:
     clients:
       all:
         - kernel_untar_build.sh
+openstack:
+  machine:
+    disk: 100 # GB
+    ram: 8000 # MB
+    cpus: 1
+  volumes: # attached to each instance
+    count: 3
+    size: 30 # GB


dachary:
<pre>
2015-10-20T20:34:26.105 INFO:tasks.ceph.osd.0.target177182.stderr:problem writing to /var/log/ceph/ceph-osd.0.log: (28) No space left on device
</pre>

suites/fs/basic/clusters/openstack.yaml
```diff
@@ -0,0 +1,8 @@
+openstack:
+  machine:
+    disk: 40 # GB
```
dachary: s/40/100/ should be enough ?

dachary:
Note that adding the file here will effectively increase the resources of **all** jobs using this directory. You probably want something finer grain. Which task requires that many resources ? You could add the **openstack:** stanza to this task only to not inflate the resource usage of every job. Precisely targeting the job is easy. By contrast, increasing the resources for all jobs is going to be difficult to fix afterward. Using the cloud means we have to worry about the price of each job to some degreee. Or we can very easily end up in a situation where 90% of the resources we pay for are actually unused and ... we have no clue how to reduce this.

dachary:
What about adding this to **tasks/cfuse_workunit_kernel_untar_build.yaml** ?

yuriw:
Trying http://pulpito.ovh.sepia.ceph.com:8081/teuthology-2015-10-20_21:32:55-fs-hammer-distro-basic-openstack/

PR #655, opened 2015-10-23
openstack: rados/thrash: allocate three disks, always

dachary:
The thrasher needs disk attached to run against xfs, ext4 or btrfs. And
some jobs use more disks and do not fit in 40GB.

http://tracker.ceph.com/issues/13450 Fixes: #13450


dachary:
I made a mistake and proposed https://github.com/ceph/ceph-qa-suite/pull/626 for hammer before doing the corresponding fix on master. This is the *forward port* of this hammer commit.

PR #656, opened 2015-10-25
fs: fix two frag_enable fragments

liewegas:
Badly formatted yaml


gregsfortytwo:
How did you spot this?

PR #660, opened 2015-10-27
added openstack confirations

yuriw:
@dachary Testing this http://pulpito.ovh.sepia.ceph.com:8081/teuthology-2015-10-27_17:12:09-ceph-deploy-firefly-distro-basic-openstack/

Signed-off-by: Yuri Weinstein yweinste@redhat.com

### DIFF
--- a/suites/ceph-deploy/fs/tasks/cfuse_workunit_suites_blogbench.yaml
+++ b/suites/ceph-deploy/fs/tasks/cfuse_workunit_suites_blogbench.yaml
@@ -22,6 +22,14 @@ roles:
   - osd.7
   - osd.8
 - - client.0
+openstack:
+  - machine:
+     disk: 40 # GB
+     ram: 8000 # MB
+     cpus: 1
+    volumes: # attached to each instance
+     count: 3
+     size: 10 # GB
 tasks:
 - install:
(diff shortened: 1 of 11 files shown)


yuriw:
@dachary seems like openstack related issues are taken care of, but we still run to other issues.  Pls take a look.

dachary:
@yuriw I think you can replace all with
<pre>
openstack:
  - machine:
     disk: 40 # GB
     ram: 8000 # MB
     cpus: 1
   volumes: # attached to each instance
     count: 3
     size: 10 # GB
</pre>
Note the dash (**-**) before machine : it is the new way of adding resources constraints and is documented at https://github.com/ceph/teuthology/#defining-instances-flavor-and-volumes.

I'm quite sure the problem you're seeing is that 10GB is not enough for the root disk: after installing the Ceph packages it already is at 8GB in most cases ;-)

yuriw:
@dachary done, thanks for comments and testing again http://pulpito.ovh.sepia.ceph.com:8081/teuthology-2015-10-28_00:01:42-ceph-deploy-firefly-distro-basic-openstack/

dachary:
@yuriw I see the test failed, did you run more of them ?

PR #662, opened 2015-10-28
use ceph-deploy for rados cls workunit in smoke

vasukulkarni:


vasukulkarni:
Logs: http://qa-proxy.ceph.com/teuthology/ubuntu-2015-10-30_11:26:27-smoke:basic-master---basic-vps/1131601/teuthology.log

vasukulkarni:
@athanatos are you ok to run raods-cls workunit using ceph-deploy instead of ceph task? I have moved error check and scrub from ceph to ceph-deploy, the above logs is from this wip branch.

vasukulkarni:
@athanatos: ping

vasukulkarni:
@alfredodeza @zmc : @athanatos is fine using ceph-deploy for this workunit, we can move independent workunits( workunits that are not dependent on daemon-helper) to use ceph-deploy task instead of ceph task as long as the conf is accurate, If you can do further review that will be good.

tasks/ceph_deploy.py
```diff
@@ -393,6 +398,50 @@ def execute_ceph_deploy(cmd):
                               'grep', '-v', 'grep', run.Raw('|'),
                               'grep', 'ceph'], check_status=False)
 
+	log.info('Checking cluster log for badness...')
+        def first_in_ceph_log(remote, pattern, excludes):
```
zmc: This indentation looks weird. Why define this function in the middle of another function?

tasks/ceph_deploy.py
```diff
@@ -393,6 +398,50 @@ def execute_ceph_deploy(cmd):
...
+                stdout=StringIO(),
+                args=args,
+                )
+            stdout = r.stdout.getvalue()
+            if stdout != '':
+                return stdout
+            return None
+
+	for remote in ctx.cluster.remotes.iterkeys():
+	    if first_in_ceph_log(remote, '\[ERR\]|\[WRN\]|\[SEC\]', config.get('log-whitelist', [])) is not None:
```
zmc: Why not simply use the result from the call(s) below?

tasks/ceph_deploy.py
```diff
@@ -574,6 +623,87 @@ def cli_test(ctx, config):
...
+                rem_site.run(args=[
+                            'sudo',
+                            'adjust-ulimits',
+                            'ceph-coverage',
+                            '{tdir}/archive/coverage'.format(tdir=testdir),
+                            'ceph', 'osd', 'deep-scrub', role])
+    prev_good = 0
+    gap_cnt = 0
+    loop = True
+    while loop:
```
zmc: Please use ``teuthology.contextutil.safe_while`` here, to avoid zombie jobs

tasks/ceph_deploy.py
```diff
@@ -618,6 +748,12 @@ def task(ctx, config):
                 stable: bobtail
              mon_initial_members: 1
              only_mon: true
+             fs: xfs|btrfs|ext4 #default xfs
+             wait-for-scrub: False
+                    #By default, the cluster log is checked for errors and warnings,
```
zmc: Please put a space after the ``#`` on these lines

tasks/ceph_deploy.py
```diff
@@ -627,6 +763,7 @@ def task(ctx, config):
         - ceph-deploy:
              branch:
                 dev: master
+             fs: xfs|btrfs|ext4 #default xfs
```
zmc: space after ``#`` please

tasks/ceph_deploy.py
vasukulkarni: @zmc I will fix this, this function was straight copy from ceph.py

tasks/ceph_deploy.py
vasukulkarni: @zmc sure I think it can be called only once.

tasks/ceph_deploy.py
vasukulkarni: This one I am not sure but will look into it, the logic is from ceph.py

tasks/ceph_deploy.py
vasukulkarni: sure

tasks/ceph_deploy.py
vasukulkarni: sure thanks.

tasks/ceph_deploy.py
vasukulkarni: That indentation is aligned with line 401, shows properly in vi. weird.

tasks/ceph_deploy.py
```diff
@@ -574,6 +619,87 @@ def cli_test(ctx, config):
...
+                            'adjust-ulimits',
+                            'ceph-coverage',
+                            '{tdir}/archive/coverage'.format(tdir=testdir),
+                            'ceph', 'osd', 'deep-scrub', role])
+    prev_good = 0
+    gap_cnt = 0
+    loop = True
+    while loop:
+	stats = get_all_pg_info(rem_site, testdir)
+        timez = [stat['last_scrub_stamp'] for stat in stats]
```
alfredodeza: @vasukulkarni there are some inconsistencies with the indentation that are showing up in more than one spot. I believe that may be because there is a mix of tabs and spaces. 

I would either switch everything to spaces for the whole file (in vim this can be done with `:set expandtab` and then `:retab 4`), or I would go through each line that has this issue. The `cat` command can help identify which ones have tabs with the following command:

    cat -e -t tasks/ceph_deploy.py

Tabs will show up like `^I`

vasukulkarni: @alfredodeza I think that could be the reason, I will run the check for this, thanks.

vasukulkarni:
Closing this for PR https://github.com/ceph/ceph-qa-suite/pull/828 which addresses the comments here and adds additional testing.

PR #663, opened 2015-10-28
tasks/cephfs: quick test for `tell`...

jcsp:
...specifically that we don't have lingering
MDS sessions after running it.  This is testing
that Client::shutdown is doing the right thing
and closing sessions.


jcsp:
Partner of https://github.com/ceph/ceph/pull/6269

PR #665, opened 2015-11-02
dummy: only use the user playbook

dachary:
Signed-off-by: Loic Dachary <ldachary@redhat.com>


dachary:
@andrewschoen would you have time to review this ? It's about reducing the run time of the dummy task (useful when testing).

dachary:
does not work apparently

PR #667, opened 2015-11-02
tasks/cephfs: simplify test_backtrace_repair

jcsp:
This was previously using a bunch of files and a small
MDCache limit to force things out of cache.  It is much
simpler to just drop the journal.

### DIFF
--- a/tasks/cephfs/test_auto_repair.py
+++ b/tasks/cephfs/test_auto_repair.py
@@ -24,48 +24,35 @@ class TestMDSAutoRepair(CephFSTestCase):
         MDS should verify/fix backtrace on fetch dirfrag
         """
 
-        # trim log segment as fast as possible
-        self.set_conf('mds', 'mds cache size', 100)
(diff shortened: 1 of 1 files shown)


tasks/cephfs/test_auto_repair.py
```diff
@@ -24,48 +24,35 @@ def test_backtrace_repair(self):
...
         self.fs.mds_asok(['flush', 'journal'])
+        self.fs.mds_stop()
 
-        self.mount_a.mount()
-        self.mount_a.wait_until_mounted()
-
-        # wait MDS to trim dentries in testdir1. 60 seconds should be long enough.
-        time.sleep(60)
+        # drop the journal so that cache will be empty on MDS restart
+        self.fs.journal_tool(["journal", "reset", "--force"])
```
ukernel: why needs this. I think 'flush journal' already trim expired journal.

jcsp: You're right, I forgot that the header is guaranteed to be written during flush.  Updated patch.

PR #669, opened 2015-11-02
client-upgrade: hammer-based test should be used on new client

dillaman:
Avoid attempting to test new release features that are not available
on older hammer OSDs.

Fixes: #13304


dillaman:
QA run: http://pulpito.ceph.com/joshd-2015-11-02_11:29:16-upgrade:client-upgrade:hammer-client-x-infernalis---basic-multi/

yuriw:
Tested http://pulpito.ceph.com/teuthology-2015-11-06_08:55:24-upgrade:client-upgrade-hammer-distro-basic-vps/

PR #678, opened 2015-11-09
suites/fs: enable all snapshost test scripts

ukernel:

### DIFF
--- a/suites/fs/snaps/tasks/snaptests.yaml
+++ b/suites/fs/snaps/tasks/snaptests.yaml
@@ -2,6 +2,20 @@ tasks:
 - workunit:
     clients:
       all:
-        - snaps/snaptest-0.sh
-        - snaps/snaptest-1.sh
-        - snaps/snaptest-2.sh
+        - fs/snaps/snaptest-0.sh
+        - fs/snaps/snaptest-1.sh
+        - fs/snaps/snaptest-2.sh
+        - fs/snaps/snap-rm-diff.sh
+        - fs/snaps/snaptest-authwb.sh
+        - fs/snaps/snaptest-capwb.sh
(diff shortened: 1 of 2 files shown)


ukernel:
please merge https://github.com/ceph/ceph/pull/6496 first

suites/fs/snaps/tasks/snaptests.yaml
```diff
@@ -2,6 +2,20 @@ tasks:
...
+        - fs/snaps/snaptest-authwb.sh
+        - fs/snaps/snaptest-capwb.sh
+        - fs/snaps/snaptest-dir-rename.sh
+        - fs/snaps/snaptest-double-null.sh
+        - fs/snaps/snaptest-estale.sh
+        - fs/snaps/snaptest-git-ceph.sh
+        - fs/snaps/snaptest-intodir.sh
+        - fs/snaps/snaptest-multiple-capsnaps.sh
+        - fs/snaps/snaptest-snap-rename.sh
+        #- fs/snaps/snaptest-snap-rm-cmp.sh
```
jcsp: Is there a tracker ticket for the reason this is commented out?  Ideally would mention in in a comment in this file

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-01-18_19:56:11-fs-greg-fs-speculative-118---basic-mira/
Only failure is known issue http://tracker.ceph.com/issues/13903

And the snapshots are nicely isolated in their own test, so it won't make noise with much else. This is step one, but before releasing them we'll also need to test against failures and fix that ceph-fuse/kernel page cache coherency issue. See http://tracker.ceph.com/issues/14427 and http://tracker.ceph.com/issues/10436.

PR #681, opened 2015-11-09
Update fio version from 2.2.9 to 2.2.11

vasukulkarni:
fio overison 2.2.11 has fix for tracker issue 13432

### DIFF
--- a/suites/krbd/rbd/tasks/rbd_fio.yaml
+++ b/suites/krbd/rbd/tasks/rbd_fio.yaml
@@ -3,7 +3,6 @@ tasks:
 - ceph: null
 - rbd_fio:
     client.0:
-       fio-version: 2.2.9
        fio-io-size: 90%
        formats: [2]
        features: [[layering],[striping]]
--- a/suites/krbd/thrash/workloads/rbd_fio.yaml
+++ b/suites/krbd/thrash/workloads/rbd_fio.yaml
@@ -1,7 +1,6 @@
 tasks:
 - rbd_fio:
     client.0:
-       fio-version: 2.2.9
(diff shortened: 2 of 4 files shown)


vasukulkarni:
@dillaman if you can merge this? It has your fix in librbd engine
Logs: http://pulpito.ceph.com/vasu-2015-11-09_11:48:47-rbd-infernalis---basic-vps/

suites/krbd/rbd/tasks/rbd_fio.yaml
```diff
@@ -3,7 +3,7 @@ tasks:
 - ceph: null
 - rbd_fio:
     client.0:
-       fio-version: 2.2.9
+       fio-version: 2.2.11
```
dillaman: I'd actually suggest yanking these overrides from all the test case YAML files and instead update the default in rbd_fio.py.

vasukulkarni: @dillaman sure I can update the default there, but we can keep this versions unchanged in yaml, it will be helpful when testing new/old fio version using overrides without any other modification.

dillaman: I am not saying remove the ability to override -- I am just saying these current "overrides" are redundant and should be removed.

vasukulkarni: @dillaman sure, i will update it in rbd_fio.py and test.

vasukulkarni:
@dillaman: with new changes http://pulpito.ceph.com/vasu-2015-11-10_11:52:53-rbd-infernalis---basic-vps/

PR #684, opened 2015-11-10
rados: 'failed to encode ...' warnings are normal on upgrades

dachary:
http://tracker.ceph.com/issues/12973 Fixes: #12973


dachary:
@liewegas does that make sense ? It showed up during a run of upgrade/hammer on **upgrade:hammer/newer/{0-cluster/start.yaml 1-install/v0.94.2.yaml 2-workload/blogbench.yaml 3-upgrade-sequence/upgrade-mon-osd-mds.yaml 4-final/{monthrash.yaml osdthrash.yaml testrgw.yaml} distros/centos_7.0.yaml} **

PR #685, opened 2015-11-10
failed to encode map e428 with expected crc

dachary:


dachary:
* *pass* http://167.114.254.108:8081/ubuntu-2015-11-12_11:34:02-upgrade:hammer-hammer-backports---basic-openstack/

PR #692, opened 2015-11-11
Added %

yuriw:
@liewegas I think it was a typo
@zmc 


liewegas:
yep, my bad!

PR #706, opened 2015-11-17
Added infernalis and jewel clients to the mix

yuriw:
@dillaman @jdurgin 

Pls review and merge

PS:  jewel client part maybe a little premature , do you think it should work?


suites/upgrade/client-upgrade/infernalis-client-x/basic/0-cluster/start.yaml
```diff
@@ -0,0 +1,14 @@
+roles:
+- - mon.a
+  - mds.a
+  - osd.0
+  - osd.1
+- - mon.b
```
dillaman: Minor: thinking about our current limited testing resources, it might be nice to have all client-upgrade tests only require two nodes: one for the OSDs/MONs and the other for the client.

yuriw: @dillaman you think something like that?

roles:
- - mon.a
  - mds.a
  - osd.0
  - osd.1
- - client.0

dillaman:
```
roles:
- - mon.a
  - mon.b
  - mon.c
  - osd.0
  - osd.1
  - osd.2
- - client.0
```

PR #707, opened 2015-11-17
Added infernalis-client-x

yuriw:
@dillaman 
@jdurgin 

This will run on jewel branch only


suites/upgrade/client-upgrade/infernalis-client-x/rbd/2-workload/rbd_notification_tests.yaml
```diff
@@ -0,0 +1,21 @@
+tasks:
+- workunit:
+    branch: infernalis
+    clients:
+      client.0:
```
dillaman: there should be client.0 and client.1 to test compatibility between the two different versions.  client.0 should be co-located w/ the OSDs/MONs and client.1 should be on its own host.

PR #718, opened 2015-11-20
rados: add test for ec fast_read

athanatos:


athanatos:
Tested, ready to merge.

athanatos:
http://pulpito.ceph.com/samuelj-2015-11-19_14:46:01-rados-infernalis-distro-basic-multi/

PR #730, opened 2015-11-23
ceph_manager: do_pg_scrub: keep scrubbing until it's done

athanatos:
The ceph pg scrub ... command isn't really guarranteed to
start a scrub, keep reissuing it until the scrub actually
happens.

Related: #12746


athanatos:
Haven't been able to get the test to fail with this.

athanatos:
http://pulpito.ceph.com/samuelj-2015-11-20_08:58:22-rados-master-distro-basic-multi/

PR #733, opened 2015-11-26
cephfs/test_client_recovery: check fsync waits for unsafe requests

ukernel:
make sure fsync wait for unsafe requests that created/modified the
file.


gregsfortytwo:
Tested in http://pulpito.ceph.com/gregf-2016-01-06_06:40:32-fs-greg-13583-test---basic-smithi/ with its ceph repo branch.

PR #737, opened 2015-11-26
suites: yaml for TestDataScan

jcsp:
This was an oversight.


jcsp:
Merge https://github.com/ceph/ceph/pull/6719 before merging this

PR #741, opened 2015-11-27
task/rbd_fio: allow for unsigned packages

dachary:
Similar to what the teuthology install.py task does, add --force-yes to
the apt-get install so that unsigned packages are successfully
installed. It is needed when the buildpackages task is used to create
packages on the fly.

There is no need to do the same for rpm packages because the
verification is controlled from the ceph-release package instead of from
the command line.

http://tracker.ceph.com/issues/13899 Fixes: #13899

(cherry picked from commit d14f2da437a3f2673aed4ea321c858f2722ac8d6)

### DIFF
--- a/tasks/rbd_fio.py
+++ b/tasks/rbd_fio.py
@@ -120,7 +120,9 @@ def run_fio(remote, config, rbd_test_dir):
         remote.run(args=['sudo', 'yum' , 'install', 'librbd1-devel', '-y'])
     elif ioengine == 'rbd':
         log.info("Installing librbd devel package on {sn}".format(sn=sn))
-        remote.run(args=['sudo', 'apt-get', '-y', 'install', 'librbd-dev'])
+        remote.run(args=['sudo', 'apt-get', '-y',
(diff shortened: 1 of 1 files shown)


dachary:
<pre>
teuthology-openstack --verbose --key-name myself --key-filename ~/Downloads/myself --suite rbd --filter='rbd/librbd/{cache/none.yaml cachepool/small.yaml clusters/fixed-3.yaml copy-on-read/off.yaml fs/btrfs.yaml msgr-failures/few.yaml workloads/rbd_fio.yaml}' --ceph-qa-suite-git-url https://github.com/dachary/ceph-qa-suite.git --suite-branch wip-13899-rbd-fio-infernalis --ceph infernalis --ceph-git-url https://github.com/ceph/ceph.git
</pre>

* *running* http://167.114.248.8:8081/ubuntu-2015-11-27_13:23:46-rbd-infernalis---basic-openstack/

Abhishekvrshny:
@dachary looks good.

PR #743, opened 2015-11-27
tasks/cephfs: add TestSessionMap.test_session_reject

jcsp:
This goes with ceph PR https://github.com/ceph/ceph/pull/6179

### DIFF
--- a/suites/fs/recovery/tasks/sessionmap.yaml
+++ b/suites/fs/recovery/tasks/sessionmap.yaml
@@ -1,4 +1,9 @@
 
+overrides:
+  ceph:
+    log-whitelist:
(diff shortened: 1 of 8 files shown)


gregsfortytwo:
This doesn't seem to be passing a (very small) integration branch, unless I've messed up somehow.

http://pulpito.ceph.com/gregf-2016-01-10_15:17:07-fs-greg-fs-testing-1-10-1---basic-mira/23357/

jcsp:
That test was run with an old version of the branch, I fixed it last week

jcsp:
Oops, actually this was missing a commit that was on wip-jcsp-testing but not here!

PR #754, opened 2015-12-04
Removed CentOS 7.1 and added CentOS 7.2 instead

yuriw:
@zmc @tmuthamizhan 

Pls review and merge
(will cherry pick to hammer, infernalis and jewel)

### DIFF
--- a/distros/all/centos_7.1.1503.yaml
+++ b/distros/all/centos_7.1.1503.yaml
@@ -1,0 +1,2 @@
+os_type: centos
+os_version: "7.1.1503"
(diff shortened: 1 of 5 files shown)


distros/all/centos_7.1.yaml
```diff
@@ -0,0 +1,2 @@
+os_type: centos
```
jdurgin: delete this one - we only need the more specific version right now

yuriw: done

PR #755, opened 2015-12-04
pg: add test cases for pg list_missing on EC pool

guangyy:
Add test case for PR: https://github.com/ceph/ceph/pull/6298

Reviewed-by: Guang Yang <yguang@yahoo-inc.com>


guangyy:
Test passed - http://qa-proxy.ceph.com/teuthology/yguang-2015-12-04_17:18:17-rados-wip-13441---basic-multi/1167683/teuthology.log

PR #763, opened 2015-12-11
buildpackages: openstack image names not unique.

robbat2:
OpenStack image names are not always unique, as noted by teuthology
commit d4c98512: openstack: image names only unique within tenants

Use the unique image ID to create the pkg host and the build hosts
instead (server create fails otherwise).

Signed-off-by: Robin H. Johnson robin.johnson@dreamhost.com

### DIFF
--- a/tasks/buildpackages/Makefile
+++ b/tasks/buildpackages/Makefile
@@ -1,6 +1,8 @@
 SHELL=/bin/bash
 D=/tmp/stampsdir
 VPATH=${D}
+TIMEOUT_SERVER_CREATE = 30m
+TIMEOUT_BUILD = 220m # 20 minutes short of 4 hours
 PKG_REPO=packages-repository
 PKG_REPO_OS_TYPE=ubuntu
 PKG_REPO_OS_VERSION=14.04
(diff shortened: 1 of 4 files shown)


tasks/buildpackages/Makefile
```diff
@@ -30,7 +37,7 @@ ${HOME}/.ssh_agent:
 	grep -q ssh_agent ~/.bashrc_teuthology || echo 'source ${HOME}/.ssh_agent' >> ~/.bashrc_teuthology
 
 flock-${PKG_REPO}:
-	openstack server create --image 'teuthology-ubuntu-14.04' --flavor ${HTTP_FLAVOR} --key-name teuthology --security-group teuthology --property ownedby=${MY_IP} --user-data ${PKG_REPO_USER_DATA} --wait ${PKG_REPO}
+	openstack server create --image ${IMAGE_ID} --flavor ${HTTP_FLAVOR} --key-name teuthology --security-group teuthology --property ownedby=${MY_IP} --user-data ${PKG_REPO_USER_DATA} --wait ${PKG_REPO}
```
dachary: this is going to incorrectly map to teuthology-centos-7.0 if called to build a centos-7.0 package, it needs to be teuthology-ubuntu-14.04 instead

robbat2: Updated PR

PR #767, opened 2015-12-15
tasks: add TestSessionMap.test_[mount|tell]_conn_close

jcsp:
To check that teardown is happening correctly when
sending commands to an MDS.

Fixes: #14048

### DIFF
--- a/tasks/cephfs/test_sessionmap.py
+++ b/tasks/cephfs/test_sessionmap.py
@@ -1,4 +1,4 @@
-
+from StringIO import StringIO
 import json
 import logging
 from tasks.cephfs.cephfs_test_case import CephFSTestCase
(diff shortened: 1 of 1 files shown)


jcsp:
Related patch: https://github.com/ceph/ceph/pull/6933

gregsfortytwo:
The commits on this branch look a little weird with the merges, which ones are supposed to be in it?

jcsp:
Don't know what happened there.  Repushed this with just the right commit.

jcsp:
Here it is running http://qa-proxy.ceph.com/teuthology/jspray-2016-01-13_01:39:24-fs:recovery-jewel---basic-mira/26640/teuthology.log

gregsfortytwo:

PR #778, opened 2015-12-24
Comments On RBD Suite Submodules.

sivaramaprasad:
Hi, Jdurgin. As you said I squash down to a single commit. Please look into my patch now.


liewegas:
@dillaman ?

dillaman:
I think this is most likely out-of-date and the preferred style is to inline the test description within the YAML via "meta: desc: <description>" blocks.

PR #780, opened 2016-01-04
tasks/cephfs: fix race in test_damage

ukernel:
When stray directory inodes are corrupted, MDS may go to damaged state
after becoming active. (MDCache::open_root/populate_mydir is called by
MDSRank::starting_done).

Fixes: #14196


gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-01-06_06:40:32-fs-greg-13583-test---basic-smithi/

PR #784, opened 2016-01-06
tasks/cephfs: update vstart_runner for teuthology change

jcsp:
Previously teuthology read test path from ctx.teuthology_config,
now it reads it from the global teuthology.config.config object.

This was breaking test_journal_migration because it uses tasks/workunit,
when run with vstart_runner.


gregsfortytwo:

PR #785, opened 2016-01-07
tasks/cephfs: fix wait_for_state timing

jcsp:
Return actual elapsed wallclock time instead of
the number of times we polled.

Fixes: #14257


jcsp:
Works running test_reconnect_timeout on a vstart cluster.

PR #786, opened 2016-01-11
tasks/cephfs: better message from test_full_fsync

jcsp:
When it gives an error from the unexpected place,
say so.


gregsfortytwo:

PR #787, opened 2016-01-11
Wip 13903 tests

jcsp:


jcsp:
DNM until we're done with #13903

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-02-17_15:11:35-fs-master---basic-mira/

PR #788, opened 2016-01-11
DNM - Removed f-h-x-offline, tiering and v0-94-4-stop tests from runs on…

yuriw:
…jewel release

@liewegas @jdurgin @tmuthamizhan @athanatos 

Pls validate my assumption that those tests do not apply to jewel release, review and merge


liewegas:
I think they do.  The problem is that debian will not ship a stable hammer, and hammer is a required upgrade from firefly -> (infernalis or jewel). They will eventually ship jewel in their next stable release.  We need to ensure that you can still upgrade by simply stopping all the firefly daemons and then starting the jewel ones.

liewegas:
same goes for v0-94-4 (keep it).  not sure what the tiering test was for.

yuriw:
@sage - ack I will close this PR then after checking on tireing part.  Can those two run on distors (centos + trusty)?

liewegas:
Pretty sure, yep

PR #789, opened 2016-01-11
tasks/cephfs: check status asok output while in reconnect

jcsp:
Mostly checking the output from the status asok is awkward
because you need to get the system stuck in a particular
state to do so.  However, we already have a test here
that sticks the system in reconnect, so here's some
very light test coverage for that asok.


gregsfortytwo:
This isn't blocked on anything else, is it? Have you run it through any testing for correctness?

jcsp:
I've run it locally

jcsp:
(and it passes)

gregsfortytwo:
http://pulpito.ceph.com/gregf-2016-02-17_15:11:35-fs-master---basic-mira/

PR #794, opened 2016-01-13
suites/rbd: avoid redundant helgrind tests

dillaman:
Allow rbd_fsx to accept an override for which valgrind tool to use.


dillaman:
Verified: 
http://qa-proxy.ceph.com/teuthology/jdillaman-2016-01-12_21:10:10-rbd-wip-14163-jewel---basic-multi/26188/teuthology.log
http://qa-proxy.ceph.com/teuthology/jdillaman-2016-01-12_21:10:10-rbd-wip-14163-jewel---basic-multi/26194/teuthology.log

PR #795, opened 2016-01-13
task/scrub_test: tests for list inconsistent obj APIs

tchaikov:


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-02-22_03:13:31-rados:basic-wip-kefu-testing---basic-multi/

PR #803, opened 2016-01-18
rbd: rbd-nbd can only run on Ubuntu

dillaman:
NBD support is not available in RHEL/CentOS 7


dillaman:
http://pulpito.ceph.com/jdillaman-2016-01-18_16:23:20-rbd-jewel---basic-multi/

PR #804, opened 2016-01-19
Added v0.94.3 to the mix

yuriw:
@liewegas @tmuthamizhan Pls review and merge

### DIFF
--- a/suites/upgrade/hammer/newer/1-install/v0.94.3.yaml
+++ b/suites/upgrade/hammer/newer/1-install/v0.94.3.yaml
@@ -1,0 +1,10 @@
+tasks:
+- install:
+    tag: v0.94.3
+- print: "**** done v0.94.3 install"
+- ceph:
+   fs: xfs
+- parallel:
+   - workload
+   - upgrade-sequence
+- print: "**** done parallel v0.94.3"


tmuthamizhan:
the comment for this PR should be "added v0.94.3 and v0.94.5 to the mix" [please note the version numbers mismatch]
also, the extension of the yaml files should be ".yaml" and NOT ".yml"

yuriw:
@tmuthamizhan fixed typos thx for good catch

tmuthamizhan:
ok, one more thing is that - latest hammer is v0.94.5, so you dont have to add that to the mix as that would be upgrading from v0.94.5 to v0.94.5 itself, instead you should probably add v0.94.4

yuriw:
@tmuthamizhan 
I think that latest hammer is not always equal to the last point release v0.94.5, so suggest to keep it, but will remove if forced :) 
And we have already v0.94.4
As of now we have:  `v0.94.2 v0.94.3  v0.94.4  v0.94.5`

yuriw:
If you think about it then upgrading from `latest giant` to `v0.94.2 v0.94.3 v0.94.4 v0.94.5` should always pass, because there is no changes in the code are introduced. 
So may get by with upgrading `v0.94.5 -> hammer`?!
Thinking aloud about usefulness of those type of suites, @tmuthamizhan @liewegas  pls comment.

PR #806, opened 2016-01-29
Made disk size in openstack 40GB, so osd won't run out of space

yuriw:
@dachary Pls review and merge

So jobs as in this run http://pulpito.ovh.sepia.ceph.com:8081/teuthology-2016-01-28_17:02:01-upgrade:hammer-x-infernalis-distro-basic-openstack/ won't fail.

(maybe we need to move it to the roles faced?)


dachary:
since it's a change that applies to the cluster, it would make more sense to have it in the cluster directory. For instance an openstack.yaml similar to rbd/qemu/clusters/openstack.yaml. It is not necessary to contain any information other than **disk: 40 # GB**

PR #807, opened 2016-01-29
suites/rados: test random queue options

liewegas:


athanatos:
I think that should be right.

PR #822, opened 2016-02-12
[DNM]: Install qemu packages required using ansible

vasukulkarni:
Install the required dependency during the start of the test


liewegas:
@vasukulkarni do we still need this?

PR #824, opened 2016-02-18
Test updates for multi-filesystem support

jcsp:
Partner of https://github.com/ceph/ceph/pull/6953


gregsfortytwo:
@jcsp, can you validate the merge in greg-fs-testing-3-7-multifs is okay? I didn't really look at what I was doing, although it seems to have passed tests.

jcsp:
Looks sane, assuming you're just talking about Filesystem.are_daemons_healthy, where the meaningful change is from wip-damage-table, and the conflict from wip-multi-filesystems is just a log line.

smithfarm:
@jcsp, @gregsfortytwo How does this not break upgrade suites? When I run, for example, `upgrade/infernalis-x` with `--suite-branch master`, it is broken due to this PR.

gregsfortytwo:
On Tuesday, March 15, 2016, Nathan Cutler <notifications@github.com
<javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:

> @jcsp <https://github.com/jcsp>, @gregsfortytwo
> <https://github.com/gregsfortytwo> How does this not break upgrade
> suites? When I run, for example, upgrade/infernalis-x with --suite-branch
> master, it is broken due to this PR.
>

We added new functionality and tests for that. If it's breaking upgrade
suites, how are we supposed to deal with it? It's not like we can never add
and test new commands to the system. :)

smithfarm:
@gregsfortytwo The upgrade suites need to be fixed, presumably. I run a test suite that is supposed to provide validation that the infernalis -> jewel upgrade path works, and tests fail. Yet this failure is not indicative of a problem with the infernalis -> jewel upgrade path.

jcsp:
@smithfarm we already have tickets for fixing this: #15049, #15124, and #15106 that you mentioned above.  The filesystem tests in general will use the new commands freely, but there are certain paths (namely the calls from ceph.py itself) that need to be backward compatible.

PR #828, opened 2016-02-22
selinux testsuite to test on CentOS

vasukulkarni:
selinux testsuite to test on CentOS, there are some known issues which on the ceph side and this pr also depends on https://github.com/ceph/teuthology/pull/790


liewegas:
@vasukulkarni needs rebase.. otherwise should we merge this?

vasukulkarni:
@liewegas closing and deleting branch, bit old,  will add 1 or 2 test directly in smoke suite.

PR #836, opened 2016-02-25
fix ceph-deploy test

vasukulkarni:
Fix issues listed here http://tracker.ceph.com/issues/14813

### DIFF
--- a/clusters/fixed-1.yaml
+++ b/clusters/fixed-1.yaml
@@ -2,7 +2,9 @@ overrides:
   ceph-deploy:
     conf:
       global:
-        mon pg warn min per osd: 1
-        osd pool default size: 1
+        osd pool default size: 2
+        osd crush chooseleaf type: 0
+        osd pool default pg num:  128
+        osd pool default pgp num:  128
 roles:
 - [mon.a, osd.0, osd.1, osd.2, client.0]
--- a/clusters/fixed-1.yaml
+++ b/clusters/fixed-1.yaml
@@ -1,2 +1,8 @@
+overrides:
+  ceph-deploy:
+    conf:
+      global:
+        mon pg warn min per osd: 1
+        osd pool default size: 1
 roles:
-- [mon.a, mon.b, mon.c, osd.0, osd.1, osd.2, client.0]
+- [mon.a, osd.0, osd.1, osd.2, client.0]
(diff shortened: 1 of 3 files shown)


vasukulkarni:
On ubuntu: http://pulpito.ceph.com/vasu-2016-02-25_10:48:34-smoke:1node-jewel---basic-mira
On CentOS: http://pulpito.ceph.com/vasu-2016-02-25_11:02:44-smoke:1node-jewel---basic-mira 
Known issue with CentOS where stop doesn't stop ceph and cause other failures, there is a tracker for this issue.

yuriw:
@vasukulkarni I assume you will want to cherry-pick it to all branches, yes?

vasukulkarni:
@yuriw sure we can backport this if needed if its failing for same reason in older release as well

dmick:
Was there a reason you didn't address the "fixed 4 seconds" or the "parsing text rather than using JSON output" aspects of the bug?

vasukulkarni:
@dmick: regarding "4 seconds" I dont think there is any issue related to that or parsing, so there is nothing to fix there, Its the default values that doesn't allow it to reach HEALTH_WARN state which has been fixed with commit a738

dmick:
Not testing for success is a bug in the test.   Fixed timeouts are wrong and cause false negatives, as this suite did in the run that made me look at it.  This PR is incomplete as it stands.

yuriw:
Please us `teuthology-describe-tests` to annotate tests

vasukulkarni:
@dmick I think there is quite a disconnect on how you understand this test, the timeout has nothing to do with status check, there are lots of status check thats happening at every cli exectuion( with by default check_status=True) and also with HEALTH_OK check. Have you seen the output from the logs i pasted?

vasukulkarni:
@dmick removed fixed timeout and also removed unnecessary overrides config items.

vasukulkarni:
@dmick I fixed all your comments, but can't see them anymore :(  probably squash removed your comment, but haven't seen before on other PRs.

clusters/fixed-1.yaml
```diff
@@ -1,2 +1,8 @@
+overrides:
+  ceph-deploy:
+    conf:
+      global:
+        mon pg warn min per osd: 1
+        osd pool default size: 1
```
dmick: These two override settings mean that there's no need to have three OSDs.  Better would be to set osd chooseleaf type 0, so that replication works as expected, along with perhaps osd pool default pg/pgp num so the pools get created large enough (don't know if the default is too small or not).  That would be testing more like a real configuration.

PR #837, opened 2016-02-26
suites: debuginfo: true for valgrind

liewegas:
This lets us get line numbers in the valgrind xml output.

### DIFF
--- a/suites/rbd/valgrind/validator/memcheck.yaml
+++ b/suites/rbd/valgrind/validator/memcheck.yaml
@@ -1,4 +1,8 @@
 overrides:
+  install:
+    ceph:
+      flavor: notcmalloc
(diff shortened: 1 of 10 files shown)


suites/rbd/valgrind/validator/memcheck.yaml
```diff
@@ -1,4 +1,8 @@
 overrides:
+  install:
+    ceph:
+      flavor: notcmalloc
```
zmc: Did you mean to add notcmalloc here?

liewegas: Yeah.  I'm surprised this test was working without it.. maybe it's being run on client code only?  In any case, doesn't hurt.

PR #842, opened 2016-02-29
rbd: updated tests to use new rbd default feature set

dillaman:


suites/rbd/cli/features/journaling.yaml
```diff
@@ -0,0 +1,6 @@
+overrides:
```
jdurgin: looks like we no longer need layering.yaml or none.yaml in this directory - and maybe a format 1 fsx or librbd api test just to catch regressions

jdurgin: these changes are all good though

dillaman: Good point

PR #846, opened 2016-03-01
cram: pin to version 0.6 because 0.7 handles \r differently

dachary:


dachary:
<code>teuthology-suite --priority 101 --suite rbd --filter="rbd/singleton/{all/formatted-output.yaml}" --suite-branch wip-14931-cram --email loic@dachary.org --ceph jewel --machine-type smithi,mira</code>

* *running* http://pulpito.ceph.com/loic-2016-02-29_23:10:38-rbd-jewel---basic-multi/

PR #853, opened 2016-03-03
suites: invoke libcephfs python tests

jcsp:
Depends on https://github.com/ceph/ceph/pull/7901

### DIFF
--- a/suites/fs/basic/tasks/libcephfs_python.yaml
+++ b/suites/fs/basic/tasks/libcephfs_python.yaml
@@ -1,0 +1,5 @@
+tasks:
+- workunit:
+    clients:
+      client.0:
+        - fs/test_python.sh


gregsfortytwo:
@jcsp, tried this in http://pulpito.ceph.com/gregf-2016-03-09_00:11:22-fs-greg-fs-testing-38-1---basic-mira/ and it seems to have failed entirely (errno 39, ENOTEMPTY...on directory cleanup, maybe?).

jcsp:
Hmm, should work if we remove the fuse task; the python tests work in the root of the filesystem, so they were tripping up on the client.0/ folder that the workunit task creates.

Updated.

gregsfortytwo:
Still busted (pretty sure I built the branch properly): http://pulpito.ceph.com/gregf-2016-03-10_23:53:12-fs-greg-fs-testing-310-1---basic-mira/

jcsp:
updated https://github.com/ceph/ceph/pull/7901 to hopefully fix this

jcsp:
Passing here: http://qa-proxy.ceph.com/teuthology/jspray-2016-03-15_08:58:50-fs:basic-wip-pycephfs-flock---basic-mira/62101/teuthology.log

PR #854, opened 2016-03-03
distros: add yaml fragments for openSUSE 13.2 and 42.1

smithfarm:
Signed-off-by: Nathan Cutler ncutler@suse.com

### DIFF
--- a/distros/all/opensuse_13.2.yaml
+++ b/distros/all/opensuse_13.2.yaml
@@ -1,0 +1,2 @@
+os_type: opensuse
+os_version: "13.2"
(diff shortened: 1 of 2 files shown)


dachary:
@smithfarm I tend to call these "yaml files" or "files". But there is a temptation to give those files names, I won't deny it. In the teuthology parlance it's "facet" :-)

dachary:
@smithfarm could you please cherry-pick this to the infernalis & hammer branch ? We're not very disciplined when it comes to backporting ceph-qa-suite and you can skip the tracker.ceph.com conventions that we normally use. If you chose to create the correpsonding issues / backports, it's fine with me too :-)

PR #859, opened 2016-03-09
Add packages.yaml

zmc:


zmc:
This is a companion to https://github.com/ceph/teuthology/pull/814

PR #860, opened 2016-03-10
Fixes for krbd test cases

dillaman:


dillaman:
Example failure: http://teuthology.ovh.sepia.ceph.com/teuthology/teuthology-2016-03-07_19:06:01-krbd-master-testing-basic-openstack/11386/teuthology.log

PR #861, opened 2016-03-10
tasks/cephfs: vstart & fuse improvements

jcsp:
A couple of corner cases that would get tripped up sometimes
in a vstart_runner.py environment.


gregsfortytwo:
Well, they haven't broken sepia teuthology tests, so good enough.

PR #865, opened 2016-03-11
packages: update for all the new debug packages we have in jewel

gregsfortytwo:


gregsfortytwo:
Passed a test in http://pulpito.ceph.com/gregf-2016-03-10_17:26:59-fs-wip-sage-testing---basic-mira/

PR #872, opened 2016-03-14
rbd: qemu live migrate

vasukulkarni:
@jdurgin @dillaman please provide initial comments and what other options we want to test

Initial runs look good on centos, I will add ubuntu support later on.
http://pulpito.ceph.redhat.com/vasu-2016-03-14_13:39:18-rbd:qemu-live-migration-jewel---basic-clara/


dillaman:
Why do you need the ```install_rbd_dependency``` task?  QEMU should already be covered in the ansible playbooks and if libvirt isn't, you probably should just update the playbook(?)

dillaman:
We'll need the VM to be running an IO workload.  It would be great if live-migration was just an attribute to the existing RBD QEMU tests (e.g. live migrate an xfstest on a cloned VM image).  It might also be good to keep live-migrating the VM back-and-forth between hosts every minute or so to really stress the test.

vasukulkarni:
@dillaman I can modify the image to have IO workload running continuously and also do migration back and forth to stress it out. In qemu xfs tests we dont use libvirt and the number of clients differ, I will look into how to use the same rbd create function so that we can run on cloned rbd image.

liewegas:
@vasukulkarni @dillaman status on this?

vasukulkarni:
Going to rework on this and resubmit.

PR #874, opened 2016-03-15
Jewel 15008

ukernel:


tasks/cephfs/test_client_limits.py
```diff
@@ -169,6 +168,19 @@ def test_client_cache_size(self):
         if not isinstance(self.mount_a, FuseMount):
             raise SkipTest("Require FUSE client to inject client release failure")
 
+        if mount_subdir:
+            # fuse assigns a fix inode number (1) to root inode. But in mounting into
+            # subdir case, the actual inode number of root is not 1. This mismatch
+            # confuses fuse_lowlevel_notify_inval_entry() when invalidating dentries
+            # in root directory.
+            self.mount_a.run_shell(["mkdir", "subdir"])
+            self.mount_a.teardown()
```
jcsp: `umount_wait` is more appropriate than `teardown` here: teardown is only needed if it is possible that something crashed.

jcsp:
Reviewed-by apart from the one small comment

PR #875, opened 2016-03-15
tasks/cephfs: support old mdsmap command during setup

jcsp:
While Filesystem at large requires the new commands, for
use from the `ceph` task we must support old style commands,
as the ceph task is used to instantiate old clusters during
upgrade testing.

Fixes: #15124, #15049, #15106

### DIFF
--- a/suites/rados/singleton-nomsgr/all/cache-fs-trunc.yaml
+++ b/suites/rados/singleton-nomsgr/all/cache-fs-trunc.yaml
@@ -6,9 +6,9 @@ tasks:
 - exec:
     client.0:
     - ceph osd pool create data_cache 4
-    - ceph osd tier add data data_cache
+    - ceph osd tier add cephfs_data data_cache
     - ceph osd tier cache-mode data_cache writeback
-    - ceph osd tier set-overlay data data_cache
+    - ceph osd tier set-overlay cephfs_data data_cache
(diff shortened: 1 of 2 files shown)


liewegas:
tox sez

./tasks/cephfs/filesystem.py:380:34: F821 undefined name 'errno'
./tasks/cephfs/filesystem.py:406:46: F821 undefined name 'errno'

otherwise lgtm

jcsp:
tox is correct!  I'm having a quick check that this really works here: http://pulpito.ceph.com/jspray-2016-03-15_06:16:01-rados:singleton-nomsgr-jewel---basic-mira

jcsp:
Passing here (along with #880):
http://pulpito.ceph.com/jspray-2016-03-15_10:04:24-rados:singleton-nomsgr-jewel---basic-mira/

gregsfortytwo:
Yep, I had it in the (all-green!) http://pulpito.ceph.com/gregf-2016-03-15_04:07:43-fs-greg-fs-testing-314---basic-mira/

And in upgrade testing with the qa-suite updates it did...better than zero, which is apparently the current state of master (http://pulpito.ceph.com/gregf-2016-03-15_12:49:59-upgrade:infernalis-x-greg-fs-testing-314---basic-mira)

PR #877, opened 2016-03-15
workunit: git-reset --hard does not accept a branch

dachary:
Uses git clone --branch refspec because it agrees to use either a tag or
a branch.


smithfarm:
Tested with both a branch and a tag - works great.

`Reviewed by: Nathan Cutler <ncutler@suse.com>`

smithfarm:
Note that the original code was not working when a branch was given, such as in https://github.com/ceph/ceph-qa-suite/blob/master/suites/upgrade/infernalis-x/parallel/2-workload/rados_api.yaml#L7

PR #880, opened 2016-03-16
suites: remove unused mdss from RADOS tests

jcsp:
These daemons were apparently unused in tests that
created an MDS daemon but never use a fuse task
to access it.


jcsp:
Passing here (along with https://github.com/ceph/ceph-qa-suite/pull/875):
http://pulpito.ceph.com/jspray-2016-03-15_10:04:24-rados:singleton-nomsgr-jewel---basic-mira/

PR #885, opened 2016-03-16
rados/singleton-nomsgr: run singleton upgrade tests on ubuntu

liewegas:


vasukulkarni:
:+1:

PR #887, opened 2016-03-16
tasks/ceph.py: Remove *.pid at end of run

dmick:
Fixes: #15162


zmc:
I'd really rather use full URLs to tickets for various reasons. Is the "Fixes: #15162" redmine magic? If so, maybe we can make it work for URLs...

dmick:
It is redmine magic if enabled, and the convention for ceph.git.  You want the full redmine URL in the commit message?

zmc:
Yes, thanks!

PR #891, opened 2016-03-17
Wip rgw new multisite

oritwas:
Changes needed for support new multisite implementation.

### DIFF
--- a/suites/rgw/multifs/overrides.yaml
+++ b/suites/rgw/multifs/overrides.yaml
@@ -3,3 +3,5 @@ overrides:
     conf:
       client:
         debug rgw: 20
+  rgw:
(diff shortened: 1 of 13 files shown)


yuriw:
@oritwas when we run this suite on jewel will `force-branch: ceph-master` have to be used as well?

oritwas:
@yuriw at the moment yes, the jewel branch fails. 
In the future we will update the jewel branch and remove this constrain

cbodley:
the multisite configuration steps look correct for both single- and multi-cluster setups

cbodley:
@oritwas is on vacation this week, but I'd like to push this forward - any objections to merging?

PR #893, opened 2016-03-18
suite/rbd: test dynamic features and rebuild object map

dillaman:


dillaman:
Associated workunits: https://github.com/ceph/ceph/pull/8216

PR #895, opened 2016-03-19
suites/rbd: small cleanup and add ceph_test_rbd_mirror workloads

jdurgin:


jdurgin:
uses https://github.com/ceph/ceph/pull/8221

dillaman:
Do you think the ceph_test_rbd_mirror could be added to the singleton test directory instead of librbd --which will repeat the test under lots of potentially unnecessary scenarios?

jdurgin:
Sure, moved it to singleton from librbd. We can worry about more coverage in a separate mirroring suite once teuthology is more capable.

I do think we'll want elements like thrashing, cache pools, and different rbd cache settings, since they do catch bugs (either in the rbd layer, or sometimes the rados layer that is only used in certain ways by rbd). With a larger suite, scheduling subsets of the suite works out well to balance coverage and resource use/day.

PR #903, opened 2016-03-21
tasks: fix ceph_deploy call to Filesystem()

jcsp:
Fixes: #15204


vasukulkarni:
thanks @jcsp

PR #904, opened 2016-03-21
upgrade: use infernalis instead of hammer for rbd tests

dachary:
There are only few differences at the moment but it may be confusing
later on if a bug is fixed in the infernalis workunits later on.

### DIFF
--- a/suites/upgrade/infernalis-x/stress-split/5-workload/rbd-cls.yaml
+++ b/suites/upgrade/infernalis-x/stress-split/5-workload/rbd-cls.yaml
@@ -3,7 +3,7 @@ meta:
    run basic cls tests for rbd
 tasks:
 - workunit:
-    branch: hammer
+    branch: infernalis
     clients:
       client.0:
         - cls/test_cls_rbd.sh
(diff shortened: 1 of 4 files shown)


dillaman:
I always thought the purpose of these tests were to ensure that old versions of librbd still function with a newer OSD (i.e. we didn't break something in the RBD class methods or the OSD didn't change semantics).

dachary:
@dillaman it's also my understanding: in the jewel branch of ceph-qa-suite, the infernalis-x verifies that installing infernalis and upgrading to the latest jewel while using librbd from infernalis works as expected. The workunit of the jewel branch should be used for that purpose, I think. Not the workunit from the hammer branch. Or am I missing something ?

jdurgin:
@dachary I think it should be the workunit from the infernalis branch for this suite, since it's testing the original client continues to behave as the cluster is upgraded. The client is never upgraded from infernalis in this suite, though a final 'upgrade client.0 and run jewel tests' might be a good addition.

yuriw:
@dachary it'd be great to use `teuthology-describe-tests` to add descriptions for cases like this for future reference :+1:

dachary:
@yuriw there are descriptions already :-)

dachary:
@jdurgin @dillaman updated and repushed with s/hammer/infernalis/ as suggested.

PR #908, opened 2016-03-25
archs: add files for each known architecture

ghost:
So that they can be prepend to teuthology-suite to run a job on a
designated architecture.

Signed-off-by: Loic Dachary loic@dachary.org


smithfarm:
@dachary Regarding d4e3cec in this PR, do you remember why it was included? (It was in the original dummy suite PR #664 but was reverted by #757, only to be reintroduced by the present PR.) 

ISTR that the user.yaml override breaks the dummy suite in OVH.

PR #911, opened 2016-03-28
Add dump_ops_enable and optrack_toggle_delay options

dzafman:
Add dump_ops_enable which continuously dumps ops using 3 commands
Add optrack_toggle_delay to alternate op tracking enablement

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -927,8 +927,8 @@ class CephManager:
             check_status=False
         ).exitstatus
 
-    def osd_admin_socket(self, osd_id, command, check_status=True):
-        return self.admin_socket('osd', osd_id, command, check_status)
+    def osd_admin_socket(self, osd_id, command, check_status=True, timeout=0):
+        return self.admin_socket('osd', osd_id, command, check_status, timeout)
 
     def find_remote(self, service_type, service_id):
         """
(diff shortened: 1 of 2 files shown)


dzafman:
@liewegas Not sure if this should be on the Jewel branch instead of master.  This tests ceph#8044 which passed my testing using wip-zafman-testing.

dzafman:
While this was working it didn't pass flake8, so I fixed that and just to be sure I'm running a single job to test.

dzafman:
@liewegas @athanatos Ready to merge unless it needs to be rebased moved to Jewel branch.

athanatos:
I think sjust@teuthology:/a/dzafman-2016-03-22_02:15:38-rados:thrash-wip-zafman-testing---basic-smithi/80901 hung waiting to join the new background task.  Presumably it's because the injectargs hung?  I think you'll have to add some kind of timeout to make sure that can't happen.

dzafman:
This must go in before ceph/ceph#7289 or tests will fail.

tasks/scrub_test.py
```diff
@@ -236,9 +236,10 @@ def _get_attrs(self, inc, attr_name):
             log.info('shard = %r' % shard)
             log.info('attr = %s' % attr_name)
             assert 'osd' in shard
-            assert attr_name in shard
             osd = shard['osd']
-            attr = shard[attr_name]
+            attr = shard.get(attr_name)
```
tchaikov: could put
```py
attr = shard.get(attr_name, False)
```

athanatos:
lgtm

PR #915, opened 2016-03-29
tasks/cephfs: update failover test for standby changes

jcsp:
Previously, an MDS could be used as a replay even
if it had "standby replay = false", this is changed
now so that replays will only happen explicitly.

### DIFF
--- a/tasks/cephfs/filesystem.py
+++ b/tasks/cephfs/filesystem.py
@@ -419,7 +419,7 @@ class Filesystem(MDSCluster):
         else:
             return False
 
-    def get_daemon_names(self, state):
+    def get_daemon_names(self, state=None):
         """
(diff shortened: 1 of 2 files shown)


jcsp:
DNM until https://github.com/ceph/ceph/pull/8300

gregsfortytwo:
@jcsp, do you think this is good to go now?
I haven't reviewed it yet but I want to make sure it shouldn't be DNM any more.

jcsp:
Yes, please go ahead and review

jcsp:
Passing tests here: http://pulpito.ceph.com/jspray-2016-05-12_04:34:36-fs-wip-jcsp-testing-20160511---basic-smithi/

@gregsfortytwo since test_standby_for_name fails without this, should probably err on the side of merging this fairly soon

PR #916, opened 2016-03-29
[DNM] tasks/cephfs: update vstart_runner for cmake build changes

jcsp:
All binaries now go into bin/


jcsp:
DNM because depends on https://github.com/ceph/ceph/pull/8351

jcsp:
This needed merging because the cmake change went in

PR #917, opened 2016-03-29
[DNM] Reproducer for #15303

jcsp:


jcsp:
DNM because will break until we fix http://tracker.ceph.com/issues/15303

jcsp:
Depends on https://github.com/ceph/ceph/pull/8353

PR #921, opened 2016-03-30
Fixes issue #15286

yuriw:
@sage pls review

whitelisted `possible circular locking dependency detected `


liewegas:
This won't work for the syslog kern.log checks.  And we decided not to whitelist it at all.  Instead, try to do most runs with -k distro, so that we only see the lockdep errors on kcephfs and krbd runs.

PR #933, opened 2016-04-05
cephfs: update tests to enable multimds when needed

gregsfortytwo:
We've disabled certain FS behaviors by default and need to turn them on when testing.


gregsfortytwo:
Goes with https://github.com/ceph/ceph/pull/8383

PR #936, opened 2016-04-06
tasks/cephfs: test health messages from readonly stat

jcsp:


jcsp:
This depends on https://github.com/ceph/ceph/pull/8468

PR #938, opened 2016-04-07
samba: run kclient mount tests on testing kernel

ukernel:
Fixes: http://tracker.ceph.com/issues/15304


ukernel:
http://pulpito.ceph.com/zyan-2016-04-07_00:46:54-samba-master---basic-multi/

PR #942, opened 2016-04-07
fs/snaps: run on xfs too, not just btrfs

liewegas:


gregsfortytwo:
Sure looks like this is removing btrfs entirely?

PR #944, opened 2016-04-07
LFN upgrade tests and a few other misc fixes

athanatos:


athanatos:
https://github.com/ceph/ceph/pull/8496

PR #953, opened 2016-04-10
buildpackages: openstack image names not unique.

ghost:
OpenStack image names are not always unique, as noted by teuthology
commit d4c98512: openstack: image names only unique within tenants

Use the unique image ID to create the pkg host and the build hosts
instead (server create fails otherwise).

Signed-off-by: Robin H. Johnson robin.johnson@dreamhost.com


dachary:
@robbat2 this is the problematic commit. I think the sane way out is to have a python helper that leverages the OpenStack* classes from teuthology to do the right thing. I'm willing to work on it tomorrow if you're running teuthology-openstack against dreamcompute. Otherwise I suppose it can wait a little.

robbat2:
@dachary Thanks; I'll think about this commit, but it's not a pressing need right now. Besides staticsites/RGW polishing, I have other commitments this week.

dachary:
@robbat2 ok. I think I know how to make this right. A single user with access to two tenants is unique to you right now, although it's the general case we should address in theory. Ff you're still using teuthology-openstack, I'll figure it out. If you're not, I'll wait until someone else runs into the same situation ;-)

liewegas:
@dachary ping?

dachary:
AFAIK teuthology-openstack is no longer maintained or used

dachary:
I'll give it a shot.

liewegas:

PR #956, opened 2016-04-11
Remove ext4 option from rados:thrash tests

vasukulkarni:
ext4 option only removed from thrash tests.


liewegas:
@athanatos ok with this?

athanatos:
@liewegas Yeah

PR #957, opened 2016-04-11
Added civetweb test case

yuriw:
@oritwas @zmc Pls review and merge


zmc:
Looks fine to me

oritwas:
@yuriw , why are we using ext4?
There are issues with ext4 and long names which effects to RGW. 
I suggest using the default fs or xfs

PR #959, opened 2016-04-12
tasks/cephfs: check data written to NS.

xiaoxichen:
NS is always used for security isolation.

Signed-off-by: Xiaoxi Chen <xiaoxchen@ebay.com>

### DIFF
--- a/tasks/cephfs/filesystem.py
+++ b/tasks/cephfs/filesystem.py
@@ -723,7 +723,7 @@ class Filesystem(MDSCluster):
             log.info("All objects for ino {0} size {1} are absent".format(ino, size))
             return True
 
-    def rados(self, args, pool=None, stdin_data=None):
+    def rados(self, args, pool=None, namespace=None, stdin_data=None):
         """
         Call into the `rados` CLI from an MDS
         """
@@ -738,7 +738,9 @@ class Filesystem(MDSCluster):
 
         # NB we could alternatively use librados pybindings for this, but it's a one-liner
         # using the `rados` CLI
-        args = [os.path.join(self._prefix, "rados"), "-p", pool] + args
+        args = ([os.path.join(self._prefix, "rados"), "-p", pool] +
+                (["--namespace", namespace] if namespace else []) +
+                args)
         p = remote.run(
             args=args,
             stdin=stdin_data,
--- a/tasks/cephfs/test_volume_client.py
+++ b/tasks/cephfs/test_volume_client.py
@@ -131,6 +131,19 @@ vc.disconnect()
         self.mounts[2].mount(mount_path=mount_path)
         self.mounts[2].write_n_mb("data.bin", 1)
 
+        #sync so that file data are persist to rados
+        self.mounts[2].run_shell(["sync"])
+
+        # Our data should stay in particular rados namespace
+        pool_name = self.mount_a.getfattr(os.path.join("volumes", group_id, volume_id), "ceph.dir.layout.pool")
+        NS_PREFIX = "fsvolumens_"
+        namespace = "{0}{1}".format(NS_PREFIX, volume_id)
+        ns_in_attr = self.mount_a.getfattr(os.path.join("volumes", group_id, volume_id), "ceph.dir.layout.pool_namespace")
+        self.assertEqual(namespace, ns_in_attr)
+
+        objects_in_ns = set(self.fs.rados(["ls"], pool=pool_name, namespace=namespace).split("\n"))
+        self.assertNotEqual(objects_in_ns, set())
+
         # De-authorize the guest
         self._volume_client_python(self.mount_b, dedent("""
             vp = VolumePath("{group_id}", "{volume_id}")


tasks/cephfs/test_volume_client.py
```diff
@@ -131,6 +131,26 @@ def test_lifecycle(self):
         self.mounts[2].mount(mount_path=mount_path)
         self.mounts[2].write_n_mb("data.bin", 1)
 
+        # Our data should stay in particular rados namespace
+        from ceph_volume_client import CephFSVolumeClient, VolumePath
```
jcsp: When these tests are run in the automated teuthology environment, the test-running machine (where this code executes) is not itself a ceph client.  For that reason, we have the "_volume_client_python" calls within these tests.  You have to wrap the actual volumeclient code in scripts for that function so that this test will work properly when run in the overnight tests.

tasks/cephfs/test_volume_client.py
```diff
@@ -131,6 +131,26 @@ def test_lifecycle(self):
...
+        vc = CephFSVolumeClient("manila", self.mount_b.config_path, "ceph")
+        vc.connect()
+        pool_name = vc._get_ancestor_xattr(vc._get_path(VolumePath(group_id, volume_id)), "ceph.dir.layout.pool")
+        ioctx = vc.rados.open_ioctx(pool_name)
+        NS_PREFIX = "fsvolumens_"
+        namespace = "{0}{1}".format(NS_PREFIX, volume_id)
+        ioctx.set_namespace(namespace)
+        it = ioctx.list_objects()
```
jcsp: See the Filesystem.rados() function which wraps the rados command line tool, this will let you run it remotely.  Use this to list objects in the namespace (similar to how it's used in Filesystem._enumerate_data_objects)

tasks/cephfs/test_volume_client.py
```diff
@@ -131,6 +131,26 @@ def test_lifecycle(self):
         self.mounts[2].mount(mount_path=mount_path)
         self.mounts[2].write_n_mb("data.bin", 1)
 
+        # Our data should stay in particular rados namespace
+        from ceph_volume_client import CephFSVolumeClient, VolumePath
+        vc = CephFSVolumeClient("manila", self.mount_b.config_path, "ceph")
+        vc.connect()
+        pool_name = vc._get_ancestor_xattr(vc._get_path(VolumePath(group_id, volume_id)), "ceph.dir.layout.pool")
```
jcsp: See Mount.getfattr for the method to read an xattr (use self.mount_a.getfattr)

jcsp:
Thanks for the test -- this just needs to be updated to use the proper remote calls instead of instantiating a VolumeClient inline, so that it will work when run in the automated environment.

xiaoxichen:
@jcsp ,thanks for the review and guidance, first time to write qa suite code:)
Could you pls review again ?

xiaoxichen:
@jcsp , i think the namespace PR to volume_client will break _enumerate_data_objects and all tests based on this function.But don't know if it's proper to read_layout before calling "rados ls"

tasks/cephfs/test_volume_client.py
```diff
@@ -131,6 +131,19 @@ def test_lifecycle(self):
         self.mounts[2].mount(mount_path=mount_path)
         self.mounts[2].write_n_mb("data.bin", 1)
 
+        #sync so that file data are persist to rados
+        self.mount_b.run_shell(["sync"])
```
jcsp: Should be self.mounts[2] instead of self.mount_b

tasks/cephfs/filesystem.py
```diff
@@ -738,7 +741,7 @@ def rados(self, args, pool=None, stdin_data=None):
 
         # NB we could alternatively use librados pybindings for this, but it's a one-liner
         # using the `rados` CLI
-        args = [os.path.join(self._prefix, "rados"), "-p", pool] + args
+        args = [os.path.join(self._prefix, "rados"), "-p", pool, "--namespace", namespace] + args
```
jcsp: This code should omit the namespace argument when it is None:
```
        args = ([os.path.join(self._prefix, "rados"), "-p", pool] +
                (["--namespace", namespace] if namespace else []) +
                args)
```

(and remove the part above where it changes None into "")

jcsp:
> @jcsp , i think the namespace PR to volume_client will break _enumerate_data_objects and all tests based on this function.But don't know if it's proper to read_layout before calling "rados ls"

To avoid breaking other users of Filesystem.rados, just omit the namespace argument when it's not set (I suggest inline how to do it).

xiaoxichen:
@jcsp fix according to your suggest.
Also update _enumerate_data_objects , to read the file_layout_t to get the namespace of the file.

jcsp:
The use of Filesystem.read_layout in the second commit is not good, because it's depending on the layout having landed in the backtrace in the data pool.  However, I don't think anything depends on that change, so can we just drop the second commit from this PR?

xiaoxichen:
@jcsp ,I did the second commit because data_objects_present and data_objects_absent( in filesystem.py) is relying on _enumerate_data_objects. If we put all data to NS, the _enumerate_data_objects will return a empty list for "exist_objects" and then fail tests in test_strays.py and test_failover.py.

Well yes it require the backlog is written, ideally we would better read them from mds?

xiaoxichen:
@jcsp dropped the second commit,

PR #967, opened 2016-04-18
Test on all supported Distros

vasukulkarni:
Test on all supported distro's

Test run:
http://pulpito.ceph.com/vasu-2016-04-16_16:15:20-krbd:rbd-jewel---basic-vps/


liewegas:
we generally get multi-distro coverage at random when this is run on smithi.. do we really need to explicitly specify the distros here?

vasukulkarni:
@liewegas  Closing and deleting the branch, doesn't seem important since we got the distro's spread evenly now.

PR #968, opened 2016-04-19
tasks/cephfs: test volume client eviction

ajarr:
Test whether the CephFSVolumeClient can evict a client based on its
auth ID and the volume path it has mounted.

Fixes: http://tracker.ceph.com/issues/15045

Signed-off-by: Ramana Raja <rraja@redhat.com>

### DIFF
--- a/suites/fs/recovery/clusters/4-remote-clients.yaml
+++ b/suites/fs/recovery/clusters/4-remote-clients.yaml
@@ -1,5 +1,5 @@
 roles:
-- [mon.a, osd.0, mds.a, mds.b, client.1, client.2]
+- [mon.a, osd.0, mds.a, mds.b, client.1, client.2, client.3]
 - [client.0, osd.1, osd.2]
 log-rotate:
   ceph-mds: 10G
--- a/tasks/cephfs/test_volume_client.py
+++ b/tasks/cephfs/test_volume_client.py
@@ -11,8 +11,8 @@ log = logging.getLogger(__name__)
 
 class TestVolumeClient(CephFSTestCase):
     # One for looking at the global filesystem, one for being
-    # the VolumeClient, one for mounting the created shares
-    CLIENTS_REQUIRED = 3
+    # the VolumeClient, two for mounting the created shares
+    CLIENTS_REQUIRED = 4
 
     def _volume_client_python(self, client, script):
         # Can't dedent this *and* the script we pass in, because they might have different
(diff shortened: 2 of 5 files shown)


tasks/cephfs/test_volume_client.py
```diff
@@ -11,8 +11,8 @@
 
 class TestVolumeClient(CephFSTestCase):
     # One for looking at the global filesystem, one for being
-    # the VolumeClient, one for mounting the created shares
-    CLIENTS_REQUIRED = 3
+    # the VolumeClient, two for mounting the created shares
+    CLIENTS_REQUIRED = 4
```
jcsp: Need to add a 4th client here: https://github.com/ceph/ceph-qa-suite/blob/master/suites/fs/recovery/clusters/2-remote-clients.yaml

(it's the hook that runs this test in the lab, might not completely make sense out of context but just need to add a client.3 after the client.1,client.2)

vasukulkarni: probably a different file with name 3-remote-clients is better?

ajarr: Yeah, did not completely understand how these hooks work. Did a union of the two above comments, and referred to https://github.com/ceph/ceph-qa-suite/commit/b06818cf4be2f822e5737bcebc089bf5dd8fc7d3 . Hope this works.

ajarr:
Depends on https://github.com/ceph/ceph/pull/8602

tasks/cephfs/test_volume_client.py
```diff
@@ -296,3 +296,109 @@ def test_15303(self):
...
+            self._volume_client_python(self.mounts[1], dedent("""
+                vp = VolumePath("{group_id}", "{volume_id}")
+                vc.deauthorize(vp, "{guest_entity}")
+                vc.delete_volume(vp)
+                vc.purge_volume(vp)
+            """.format(
+                group_id=group_id,
+                volume_id=volume_ids[i],
+                guest_entity=guest_entity
+            )))
```
jcsp: I think you also need ```self.mounts[2].umount_wait(force=True)``` at the end here, as otherwise I found that the umount during teardown got stuck.

suites/fs/recovery/clusters/4-remote-clients.yaml
```diff
@@ -0,0 +1,6 @@
+roles:
```
jcsp: Also remove the 2-remote-clients.yaml file, otherwise all the tests will run twice, once with the old file and once with the new.  The teuthology test runner treats the yaml files in each dir as a list of permutations to apply to the tests.

jcsp:
I'm running a modified version[1] of this overnight in the lab, will post link in morning.

1.https://github.com/ceph/ceph-qa-suite/commits/wip-15045-jcsp

suites/fs/recovery/clusters/4-remote-clients.yaml
```diff
@@ -0,0 +1,6 @@
...
+- [mon.a, osd.0, mds.a, mds.b, client.1, client.2, client3]
```
jcsp: s/client3/client.3/

jcsp:
Passing here: http://pulpito.ceph.com/jspray-2016-05-04_03:38:26-fs:recovery-wip-15045---basic-mira/

gregsfortytwo:
Goes with a Ceph PR

PR #970, opened 2016-04-22
Fix for 'branch' overrides to work

vasukulkarni:
teuthology default override template defines devcommit which makes
'branch' overrides in the test useless since it only picksup one option
Remove the devcommit overrides from the template and use that as
default in case no other 'branch' override exist.

### DIFF
--- a/tasks/ceph_deploy.py
+++ b/tasks/ceph_deploy.py
@@ -239,14 +239,19 @@ def build_ceph_cluster(ctx, config):
                         ceph_admin, conf_path, lines, sudo=True)
 
         # install ceph
-        install_nodes = './ceph-deploy install ' + \
-            (ceph_branch if ceph_branch else "--dev=master") + " " + all_nodes
+        ceph_sha = ctx.config['sha1']
+        devcommit = '--dev-commit={sha}'.format(sha=ceph_sha)
+        if ceph_branch:
+            option = ceph_branch
+        else:
+            option = devcommit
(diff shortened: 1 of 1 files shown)


vasukulkarni:
Depends on https://github.com/ceph/teuthology/pull/849

tasks/ceph_deploy.py
```diff
@@ -239,14 +239,19 @@ def execute_ceph_deploy(cmd):
                         ceph_admin, conf_path, lines, sudo=True)
 
         # install ceph
-        install_nodes = './ceph-deploy install ' + \
-            (ceph_branch if ceph_branch else "--dev=master") + " " + all_nodes
+        ceph_sha = ctx.config['sha1']
+        devcommit = '--dev-commit={sha}'.format(sha=ceph_sha)
+        if ceph_branch:
+            option = ceph_branch
```
alfredodeza: `devcommit` is being constructed with `--dev-commit` but `ceph_branch` isn't. Shouldn't this line be:

    option = '--dev={branch}'.format(branch=ceph_branch)

alfredodeza:
this looks OK to me

vasukulkarni:
@alfredodeza thanks, I will do one final test and merge.

vasukulkarni:
http://qa-proxy.ceph.com/teuthology/vasu-2016-06-22_13:52:51-ceph-deploy-master---basic-vps/271405/teuthology.log failure is unrelated to this change, uses stable jewel branch

PR #976, opened 2016-04-28
suites/hadoop: use xfs

liewegas:
This avoids working aroudn ext4 limits.  Whatev.


gregsfortytwo:
I was kind of hoping not to throw all our ext4 away, but....

PR #977, opened 2016-04-28
rgw/verify: use rgw_compression_type = "random"

cbodley:
~~This is my naive attempt to add `rgw_compression_type={none, snappy, zlib}` to the test matrix of the rgw suite to validate https://github.com/ceph/ceph/pull/7715. I would eventually like to narrow down which parts of the rgw suite test with compression.. maybe just running s3tests against each type is good enough.~~

This uses the new compressor plugin factory's support for the compression type "random" (see https://github.com/ceph/ceph/pull/11901) to get test coverage of all supported compression types without expanding the test matrix of the rgw suite.

### DIFF
--- a/suites/rgw/verify/overrides.yaml
+++ b/suites/rgw/verify/overrides.yaml
@@ -3,5 +3,6 @@ overrides:
     conf:
       client:
         debug rgw: 20
+        rgw compression type: random
   rgw:
     frontend: civetweb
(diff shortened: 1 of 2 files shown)


cbodley:
removed `[DNM]` now that support for `rgw_compression_type = "random"` merged into master

cbodley:
added `[DNM]` pending https://github.com/ceph/ceph/pull/12113 which moves the configuration into the zone

cbodley:
updated to use the new zone placement configuration for compression. for multisite tests, it adds this compression field directly into the json format. for non-multisite tests, it runs the new `radosgw-admin zone placement modify` command to set the compression type

cbodley:
@yehudasa @oritwas rgw:verify suite passing with only valgrind failures: http://pulpito.ceph.com/cbodley-2016-12-06_10:46:50-rgw:verify-wip-cbodley-testing---basic-smithi

oritwas:
lgtm

PR #978, opened 2016-04-28
test fast ec read

liewegas:


athanatos:
Do we want this enabled for all ec tests?

liewegas:
nope, that's why there is a none.yaml too (empty).

athanatos:
LGTM

PR #980, opened 2016-05-02
Fix infinite wait during monitor quorum check

vasukulkarni:
The current loop can cause the logs to run into huge file and loop forever in case of dead monitors, This change will cause it exit after 10minutes if they dont reach quorum.


vasukulkarni:
@athanatos @liewegas: Do you think 10minutes is sufficient for all the cases? , I am rerunning some mon thrash tests as well.

liewegas:
10m should be plenty.  rebase?

athanatos:
should be ok

vasukulkarni:
@liewegas: rebased and tested here: http://pulpito.ceph.com/vasu-2016-12-06_01:06:54-upgrade:jewel-x-master---basic-vps/

PR #985, opened 2016-05-07
suites/rados: test osd_discard_disconnected_ops

ZhenyuLeng:
In ceph.git i propose "PG: no need to judge op_is_discardable after op has been dequeued" https://github.com/ceph/ceph/pull/8717

Here to test the parameter osd_discard_disconnected_ops

### DIFF
--- a/suites/rados/thrash/workloads/readwrite.yaml
+++ b/suites/rados/thrash/workloads/readwrite.yaml
@@ -1,6 +1,9 @@
 overrides:
   ceph:
     crush_tunables: optimal
+    conf:
+      osd:
+        osd_discard_disconnected_ops: false
 tasks:
 - rados:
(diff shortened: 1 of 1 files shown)


liewegas:
you want something like
<pre>
overrides:
  ceph:
    conf:
      osd:
        osd discard disconnected ops: false
</pre>
to add/set the config option.  it's not a rados.py option...

ZhenyuLeng:
@liewegas Thank you very much for the reply!
Is this time the change is OK?

liewegas:
Yep!  Can you squash this down into one commit please?  (git rebase -i HEAD^^^^)

ZhenyuLeng:
I have done it right now ^_^ @liewegas

PR #986, opened 2016-05-09
rados: test jewel instead of legacy tunables

liewegas:
Legacy tunables semi-frequently turn up crush mappings that only
return 1 instead of 2 results.  Switch to covering jewel tunables
explicitly instead.


athanatos:
LGTM!

PR #991, opened 2016-05-12
tasks/cephfs: fix ps usage to not truncate cmds

jcsp:
Now that daemon commandlines include "--cluster" they
easily overflow ps's default behaviour.

Fixes: http://tracker.ceph.com/issues/15857


jcsp:
Tested locally (that test_sessionmap passes against vstart cluster)

PR #994, opened 2016-05-14
make-rpm.sh: support openSUSE

smithfarm:
With the patch in this PR, I can run the `buildpackages/any` suite on openSUSE Leap in OVH. Though it does not yet complete successfully, it gets all the way through the `make-rpm.sh` script and into the Ansible section (which fails because it doesn't support openSUSE yet).

### DIFF
--- a/tasks/buildpackages/make-rpm.sh
+++ b/tasks/buildpackages/make-rpm.sh
@@ -30,7 +30,14 @@ sha1=$4
 flavor=$5
 arch=$6
 
-sudo yum install -y git
+suse=false
+[[ $codename =~ suse ]] && suse=true
+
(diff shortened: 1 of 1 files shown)


smithfarm:
@dachary Before we merge, I will run this through a `buildpackages/any` targeting CentOS 7.2 so I can assert that it doesn't break existing functionality.

smithfarm:
**running** http://149.202.184.76:8081/ubuntu-2016-05-14_20:33:34-buildpackages:any-master---basic-openstack/

smithfarm:
@dachary Passes `buildpackages/any` on CentOS 7.2.

PR #1008, opened 2016-05-19
Wip systemd tests

vasukulkarni:
@branto1 @ktdreyer  Tests for systemd on both ubuntu and centos, 

Right now its failing which i believe is due to systemd issue,  will be testing on some more branches

http://pulpito.ceph.com/vasu-2016-05-19_09:44:01-smoke:systemd-jewel---basic-mira/202058/

### DIFF
--- a/tasks/systemd.py
+++ b/tasks/systemd.py
@@ -68,6 +68,7 @@ def task(ctx, config):
                              osd_service])
             remote.run(args=['sudo', 'systemctl', 'stop',
                              osd_service])
+            time.sleep(4) #immediate check will result in deactivating state
             r = remote.run(args=['sudo', 'systemctl', 'status', osd_service],
                            stdout=StringIO(), check_status=False)
             log.info(r.stdout.getvalue())
@@ -81,6 +82,7 @@ def task(ctx, config):
         if mon_role_name in roles:
             remote.run(args=['sudo', 'systemctl', 'status', mon_name])
             remote.run(args=['sudo', 'systemctl', 'stop', mon_name])
+            time.sleep(4) #immediate check will result in deactivating state
             r = remote.run(args=['sudo', 'systemctl', 'status', 'ceph.target'],
                            stdout=StringIO(), check_status=False)
             if r.stdout.getvalue().find('Active: inactive'):
@@ -92,6 +94,7 @@ def task(ctx, config):
         if mds_role_name in roles:
             remote.run(args=['sudo', 'systemctl', 'status', mds_name])
             remote.run(args=['sudo', 'systemctl', 'stop', mds_name])
+            time.sleep(4) #immediate check will result in deactivating state
             r = remote.run(args=['sudo', 'systemctl', 'status', 'ceph.target'],
                            stdout=StringIO(), check_status=False)
             if r.stdout.getvalue().find('Active: inactive'):
--- a/tasks/systemd.py
+++ b/tasks/systemd.py
@@ -1,0 +1,103 @@
+"""
+Systemd test
+"""
+import contextlib
+import logging
+import re
+import time
+
+from cStringIO import StringIO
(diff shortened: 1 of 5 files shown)


ktdreyer:
:+1:

branto1:
@vasukulkarni You are hitting

`Running: 'sudo systemctl list-units | grep ceph'`
`stderr:sudo: systemctl: command not found`

in that run which usually means that systemctl is not in PATH or is not an executable for the user for whatever reason. Are you sure the system have systemd installed? Is it in PATH for the user -- i.e. does it help if you use 'sudo /usr/bin/systemctl' instead? Is it executable for the user?

vasukulkarni:
@branto1, I actually made some changes later on, for the original error I coudn't understand why it failed but the later runs worked with no change there, also there is a issue with systemctl list-units since one of the service comes up in failed state, here is a new run on ubuntu/centos, ubuntu worked fine but looks like there is an issue with centos, I am making one minor change to handle osd numbering, will update results here.

http://pulpito.ceph.com/vasu-2016-05-23_17:50:23-smoke:systemd-master---basic-vps/

vasukulkarni:
@ktdreyer / @b-ranto  can one of you review/merge this, the issue with failed ubuntu test link which i pasted above was due to 14.04 which doesn't work with systemd, i made the changes and the test now runs with 7.2/16.04 http://pulpito.ceph.com/vasu-2016-06-17_12:41:05-smoke:systemd-jewel---basic-vps/265730/

vasukulkarni:
@ktdreyer ping, tests pass on both centos and xenial

tasks/systemd.py
```diff
@@ -0,0 +1,106 @@
...
+                             'grep', 'ceph'], stdout=StringIO(),
+                       check_status=False)
+        log.info(r.stdout.getvalue())
+        if r.stdout.getvalue().find('failed'):
+            log.info("Ceph services in failed state")
+
+        # test overall service stop and start
+        log.info("Stopping all Ceph services")
+        remote.run(args=['sudo', 'systemctl', 'stop', 'ceph.target'])
+        r = remote.run(args=['sudo', 'systemctl', 'status', 'ceph.target'],
```
ktdreyer: I think we agreed that `systemctl status ceph.target` doesn't really do anything?

vasukulkarni: @ktdreyer the status works, the one that doesn't work is the ceph-disk service that comes up in failed state for which I am ignoring the errror for now until its fixed(line 28)
http://pulpito.ceph.com/vasu-2016-06-21_14:32:12-smoke:systemd-jewel---basic-vps/

2016-06-21T17:11:01.363 INFO:teuthology.orchestra.run.vpm097:Running: 'sudo systemctl list-units | grep ceph'
2016-06-21T17:11:01.452 INFO:tasks.systemd:Ceph services in failed state

ktdreyer: What I mean is that it is not meaningful to run `systemctl status ceph.target` - it doesn't really give you any information about whether Ceph is running or not.

vasukulkarni: @ktdreyer sure I am going to remove that check and push

ktdreyer:
This looks good to me

vasukulkarni:
@ktdreyer thanks, i just added a radosbench test at the end to ensure everything works, will test it once.

tasks/systemd.py
b-ranto: You could probably use something like 'systemctl status ceph*.service' to check the state of the daemons. The systemctl command does allow *-expansions.

vasukulkarni:
http://pulpito.ceph.com/vasu-2016-06-29_21:21:43-smoke:systemd-jewel---basic-vps/

yuriw:
@vasukulkarni was the assumption after using distos that smoke will run only on vps ?

PR #1012, opened 2016-05-20
suites/teuthology/multi-cluster: make them pass again

jdurgin:
Use xfs to avoid object length limits and be explicit about which
clients to run on, so the workunit git clone doesn't race with another
client on the same machine.


jdurgin:
@dillaman need this for the teuthology:multi-cluster suite

PR #1013, opened 2016-05-23
tasks/cephfs: fix failure in test_mount_conn_close

jcsp:
Passing here: http://qa-proxy.ceph.com/teuthology/jspray-2016-05-23_11:57:19-fs:recovery-master---basic-mira/209404/teuthology.log

Fixes: http://tracker.ceph.com/issues/15900


jcsp:
Quick review @gregsfortytwo ?

gregsfortytwo:
I don't get it; why were we matching on daemon-helper in a bad way?
Oh, because daemon-helper is its own process with ceph-mds as one of the arguments, d'oh.

PR #1015, opened 2016-05-26
(jewel) tasks/cephfs: update failover test for standby changes

jcsp:
Previously, an MDS could be used as a replay even
if it had "standby replay = false", this is changed
now so that replays will only happen explicitly.

(cherry picked from commit d9309b57e3c12579f64939ee78161f7e99a062c8)

### DIFF
--- a/tasks/cephfs/test_failover.py
+++ b/tasks/cephfs/test_failover.py
@@ -365,17 +365,8 @@ class TestMultiFilesystems(CephFSTestCase):
         self.mds_cluster.mds_restart(mds_c)
         self.wait_for_daemon_start([mds_a, mds_c])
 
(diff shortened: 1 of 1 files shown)


jcsp:
Backport needed to avoid failure in test_standby_for_name http://pulpito.ceph.com/teuthology-2016-05-22_02:10:01-fs-jewel---basic-smithi/207197/

jcsp:
@dachary should this go into a test run or just merge it?

jcsp:
This was (amongst other things) on http://pulpito.ceph.com/jspray-2016-06-02_15:43:38-fs-wip-manila-backports---basic-mira/

PR #1018, opened 2016-05-26
tasks/cephfs: reproducer for #16022

jcsp:
Even though check_subs was broken, the multiclient
test was still passing because it didn't test that
clients survived a failover in the multi-filesystem
case.

Fixes: http://tracker.ceph.com/issues/16022


jcsp:
I've tested locally that this fails without "MDSMonitor.cc: fix mdsmap.<namespace> subscriptions" and passes with it.

jcsp:
Passing here http://pulpito.ceph.com/jspray-2016-05-26_17:59:25-fs-wip-jcsp-testing-20160527---basic-mira/217660

jcsp:
@ukernel could you review please?

jcsp:
@ukernel ping, please could you review this?

PR #1020, opened 2016-05-26
Added file locking logic to avoid possible collisions …

yuriw:
…when appending to the log file

@jdurgin @dmick @tmuthamizhan 

Tested with 420 seconds wait if one process locked the log (similar what would rados suite would take to run) and it seemed sufficient. 

Pls review and merge

Fixes http://tracker.ceph.com/issues/15563

### DIFF
--- a/nightlies/cron_wrapper
+++ b/nightlies/cron_wrapper
@@ -8,12 +8,11 @@ if [ -z $1 ]; then
 fi
 
 # set permanent $LOG file var
-# set per-job $LOG_XX file
-# set $LOG_LOCKED_ERR in case locking failed
 LOG="/var/log/crontab-nightlies-log/crontab.log"
+# set $LOG_LOCKED_ERR in case locking failed
 LOG_LOCK_ERR="/var/log/crontab-nightlies-log/crontab_lock_problem.$$"
 
-# temp files to store sdtout and stderr
+# temp files to store stdout and stderr
 # named with the PID of this script in their name so they'll be unique
 STDERR="/var/tmp/stderr.$$"
 STDOUT="/var/tmp/stdout.$$"
@@ -21,11 +20,12 @@ STDOUT="/var/tmp/stdout.$$"
 # $STDOUT and $STDERR are removed when the script exits for any reason
 trap  "rm -f $STDOUT $STDERR" 0
 
-# log a name of the command about to be run
-echo "Running command: $@" >> $STDOUT
-
 # run a command from this script's argument
 # redirect stdout to $STDOUT file and redirect stderr to $STDERR file
+
+DATE=$(date)
+echo -n "$DATE: "  >> $STDOUT
+echo "Running command: $@" >> $STDOUT
 "$@" > $STDOUT 2> $STDERR
 
 # get return code from the command run
@@ -39,17 +39,12 @@ if [ $code != 0 ] ; then
         cat $STDERR
 else
         # normal exit: just log stdout
-        DATE=$(date)
-        echo -n "$DATE: "  >> $STDOUT
-        cat $STDOUT   >> $STDOUT
-fi
 
-# lock $LOG with file descriptor 200
-# if $LOC is locked by other process - wait for 20 sec
-# if lock failed, write to $LOG_LOCK_ERR, e.g. all information already in $LOG_XX
-exec 200>>$LOG
-flock -w 20 200 || LOG=$LOG_LOCK_ERR
-# write from per-job $LOG_XX file to the commonn $LOG
-cat $LOG_XX >> $LOG
-# unlock 
-flock -u 200
+	# lock $LOG with file descriptor 200
+	exec 200>>$LOG
+	# if $LOG is locked by other process - wait for 20 sec
+	flock -w 20 200 || LOG=$LOG_LOCK_ERR
+	cat $STDOUT >> $LOG
+	# unlock
+	flock -u 200
+fi
--- a/nightlies/cron_wrapper
+++ b/nightlies/cron_wrapper
@@ -8,7 +8,10 @@ if [ -z $1 ]; then
 fi
 
 # set permanent $LOG file var
+# set per-job $LOG_XX file
+# set $LOG_LOCKED_ERR in case locking failed
 LOG="/var/log/crontab-nightlies-log/crontab.log"
+LOG_LOCK_ERR="/var/log/crontab-nightlies-log/crontab_lock_problem.$$"
 
 # temp files to store sdtout and stderr
 # named with the PID of this script in their name so they'll be unique
@@ -19,7 +22,7 @@ STDOUT="/var/tmp/stdout.$$"
 trap  "rm -f $STDOUT $STDERR" 0
 
 # log a name of the command about to be run
-echo "Running command: $@" >> $LOG
+echo "Running command: $@" >> $STDOUT
 
 # run a command from this script's argument
 # redirect stdout to $STDOUT file and redirect stderr to $STDERR file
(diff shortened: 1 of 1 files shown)


nightlies/cron_wrapper
```diff
@@ -18,6 +18,11 @@ STDOUT="/var/tmp/stdout.$$"
 # $STDOUT and $STDERR are removed when the script exits for any reason
 trap  "rm -f $STDOUT $STDERR" 0
 
+# lock $LOG with file descriptor 200
+# if $LOC is locked by other process - wait for 420 sec
+exec 200>>$LOG
+flock -w 420 200 || exit 1
```
dmick: 1) 420 seems weirdly arbitrary.  Why not, oh, 15 minutes?

2) if you fail to lock the log, you don't want to fail to run the job altogether; just avoid logging, probably, or perhaps fallback to "send email with all output", or "log to a fallback file"

3) you really only need to hold the lock for the time you need to write the lock.

jdurgin: logging to a fallback, perhaps with no locking on the fallback sounds good

yuriw:
@jdurgin @dmick Untested fixes per comment and discussions

nightlies/cron_wrapper
```diff
@@ -18,6 +18,13 @@ STDOUT="/var/tmp/stdout.$$"
 # $STDOUT and $STDERR are removed when the script exits for any reason
 trap  "rm -f $STDOUT $STDERR" 0
 
+# lock $LOG with file descriptor 200
+# if $LOC is locked by other process - wait for 420 sec
+exec 200>>$LOG
+# if flock failed, define a new $LOG and keep runing
+flock -w 420 200 || LOG="/var/log/crontab-nightlies-log/crontab_lock_problem.log"
+
+
```
dmick: ok, that's one of my three points, but  

1) still 420s? 

2) because you're locking before running the command, that means that the entire time the command runs, no one else can write to the log.   Say job A runs for 10s and job B for 419s.  They both start simultaneously, and then job B gets the lock on the shared log; job A could run and complete, but is blocked for the entire runtime of job B.

The solution is to lock only when writing to the shared log, and to drop the lock when not writing.  Since the command itself already outputs to separate files, you can fix this with more invocations of flock.

yuriw: @dmick incorporated comments

nightlies/cron_wrapper
```diff
@@ -37,6 +46,10 @@ if [ $code != 0 ] ; then
 else
         # normal exit: just log stdout
         DATE=$(date)
-        echo -n "$DATE: " >> $LOG
-        cat $STDOUT  >> $LOG
+        # if flock failed, define a new $LOG and keep runing
+        flock 200 || $LOG
```
dmick: What does this line do if flock fails?  (also: 'running')

yuriw: @dmick 
I think this line 50 is wrong, should be `flock 200 || LOG="/var/log/crontab-nightlies-log/crontab_lock_problem.log"`

dmick: however, if you failed to get the lock above, you probably want to continue using that fallback lock file for the rest of the job, otherwise messages will exist in both places, which would be bad.  Also also, you probably want the fallback log file name to be unique per-job (simplest method, adding $$ to the name).

nightlies/cron_wrapper
```diff
@@ -19,7 +23,7 @@ STDOUT="/var/tmp/stdout.$$"
 trap  "rm -f $STDOUT $STDERR" 0
```
dmick: this is here to remove temporary files, and needs to remove LOG_XX now

yuriw:
@jdurgin @dmick if there is no more comments, pls merge.

nightlies/cron_wrapper
```diff
@@ -37,6 +41,16 @@ if [ $code != 0 ] ; then
 else
         # normal exit: just log stdout
         DATE=$(date)
-        echo -n "$DATE: " >> $LOG
-        cat $STDOUT  >> $LOG
+        echo -n "$DATE: "  >> $LOG_XX
+        cat $STDOUT   >> $LOG_XX
```
dmick: So, given that you still have STDOUT, and given that it's private to this process, there's really no need for LOG_XX; you could just echo the Running command... and the date directly to $STDOUT and remove LOG_XX altogether

yuriw: @dmick agreed and done

nightlies/cron_wrapper
```diff
@@ -37,6 +40,16 @@ if [ $code != 0 ] ; then
...
+        cat $STDOUT   >> $STDOUT
 fi
+
+# lock $LOG with file descriptor 200
+# if $LOC is locked by other process - wait for 20 sec
+# if lock failed, write to $LOG_LOCK_ERR, e.g. all information already in $LOG_XX
+exec 200>>$LOG
+flock -w 20 200 || LOG=$LOG_LOCK_ERR
+# write from per-job $LOG_XX file to the commonn $LOG
+cat $LOG_XX >> $LOG
```
dmick: so....

yuriw: oops, fixed..

dmick: did you not push the change?

yuriw:
@dmick github is acting up, so added a fresh commit

nightlies/cron_wrapper
```diff
@@ -8,7 +8,9 @@ if [ -z $1 ]; then
 fi
 
 # set permanent $LOG file var
+# set $LOG_LOCKED_ERR in case locking failed
```
dmick: actual name is LOG_LOCK_ERR.  comment should probably go above the line setting the variable instead of above the LOG= line.

nightlies/cron_wrapper
```diff
@@ -37,6 +39,15 @@ if [ $code != 0 ] ; then
...
+        cat $STDOUT   >> $STDOUT
 fi
+
+
+# lock $LOG with file descriptor 200
+# if $LOC is locked by other process - wait for 20 sec
+exec 200>>$LOG
+flock -w 20 200 || LOG=$LOG_LOCK_ERR
+cat $STDOUT >> $LOG
+# unlock
```
dmick: git cli shows that there is a trailing space here which shouldn't be here

nightlies/cron_wrapper
```diff
@@ -37,6 +39,15 @@ if [ $code != 0 ] ; then
...
 fi
+
+
+# lock $LOG with file descriptor 200
+# if $LOC is locked by other process - wait for 20 sec
+exec 200>>$LOG
+flock -w 20 200 || LOG=$LOG_LOCK_ERR
+cat $STDOUT >> $LOG
+# unlock 
+flock -u 200
```
dmick: Does this exit without error if the lock is not held (if you failed to get it above)?

nightlies/cron_wrapper
```diff
@@ -37,6 +39,17 @@ if [ $code != 0 ] ; then
 else
         # normal exit: just log stdout
         DATE=$(date)
-        echo -n "$DATE: " >> $LOG
-        cat $STDOUT  >> $LOG
+        echo -n "$DATE: "  >> $STDOUT
+        cat $STDOUT   >> $STDOUT
```
dmick: This isn't going to work; doesn't make any sense

yuriw:
@dmick anything else left here?
@tmuthamizhan FYI

dmick:
In an attempt to bring this to a quicker close, I've added some changes in another branch, wip-15563-dmick, for discussion.  I think we lost track of the actual intent of the wrapper logging and I'm trying to quickly converge.

yuriw:
@dmick updated thx !

dmick:
You missed a bunch of the changes I made; did you disagree with them for some reason, or?...

(Some of them were things like spelling)

yuriw:
no not on purpose, will double check...

yuriw:
@dmick done

PR #1024, opened 2016-05-27
tasks/ceph: default to xfs instead of root dir

vasukulkarni:
Default to xfs instead of using root fs dir
for osd


liewegas:
lgtm, but we should do a quick test here?

PR #1026, opened 2016-05-28
scrub_test: Shards are not marked in error when osd guesses

dzafman:
This should fix the errors seen in the rados runs.


dzafman:
This run passed which would have failed without the test changes.

http://qa-proxy.ceph.com/teuthology/dzafman-2016-05-27_19:20:19-rados:basic-wip-16064---basic-smithi/219122/teuthology.log

PR #1028, opened 2016-06-01
rbd: provision volumes to format as XFS

dachary:


dillaman:
@dachary Need to update ```suite/rbd/cli/clusters```, ```suite/rbd/valgrind/clusters```, and ```suite/rbd/mirror/clusters```

PR #1031, opened 2016-06-03
tasks: fix non-existent sleep function

jdurgin:
CephManager has no sleep function. Use time.sleep() instead.

Ran into this while testing a branch. Apparently it doesn't happen
much since this hasn't changed in years, but the error was copied
into several tasks.


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-06-04_23:16:15-rados-wip-kefu-testing---basic-mira/

PR #1043, opened 2016-06-07
vstart_runner: only kill/look at user's processes

batrick:
This prevents vstart_runner.py from trying to kill other user's ceph-mds/ceph-fuse instances.


batrick:
Oh, there is a bug I think. Don't merge...

batrick:
This is ready now.

jcsp:
:+1:

PR #1045, opened 2016-06-08
tasks/cephfs: fix vstart runner _get_pid

jcsp:
Dropping the 'x' caused the ps lines not to include
the "-i <myid>" part of the command line that we use
to find named daemons' pids.


batrick:
Oops. I'll give this a try now.

batrick:
Worked for me, merged.

PR #1046, opened 2016-06-08
cephfs: test readahead is working

batrick:
Test case for: https://github.com/ceph/ceph/pull/9374
Issue: http://tracker.ceph.com/issues/16024

### DIFF
--- a/suites/fs/recovery/tasks/client-readahad.yaml
+++ b/suites/fs/recovery/tasks/client-readahad.yaml
@@ -1,0 +1,4 @@
+tasks:
+  - cephfs_test_runner:
+      modules:
+        - tasks.cephfs.test_readahead
--- a/tasks/cephfs/test_readahead.py
+++ b/tasks/cephfs/test_readahead.py
@@ -1,0 +1,26 @@
+import logging
+from tasks.cephfs.cephfs_test_case import CephFSTestCase
+
+log = logging.getLogger(__name__)
+
+class TestReadahead(CephFSTestCase):
+    def test_flush(self):
+        # Create 32MB file
+        self.mount_a.run_shell(["dd", "if=/dev/urandom", "of=foo", "bs=1M", "count=32"])
+
+        # Unmount and remount the client to flush cache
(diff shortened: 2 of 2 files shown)


batrick:
I tested this against 545dde68e1040f368ff94e012530456355c2144a (fails) and 28a9e45 (succeeds) using vstart_runner.py against a vstart instance.

gregsfortytwo:
Oh, this is a new test, which I missed. In that case we need to add it to to an appropriate "tasks" folder in the fs suite. There are examples of how to do that if you grep for cephfs_test_runner.

If there were a good one to add it to I'd prefer that since it cuts down on the number of test setups we have to go through in a run, but I'm not seeing anything obvious so I guess we can use this as the first of a new set of client-performance-functionality tests. ;)

jcsp:
Reviewed-by me.  Just needs a yaml in suites/fs/recovery/tasks as greg mentioned.

gregsfortytwo:
Hmm, the test failed in http://pulpito.ceph.com/gregf-2016-06-10_01:23:22-fs-greg-fs-testing-69-1---basic-mira/249915/

>2016-06-10T02:26:45.012 INFO:tasks.cephfs.test_readahead:read operations: 33
2016-06-10T02:26:45.036 INFO:tasks.cephfs_test_runner:test_flush (tasks.cephfs.test_readahead.TestReadahead) ... ERROR

Are we doing some setup differently between teuthology and vstart that could account for this and needs to be changed?

batrick:
Oh, the test should fail right? This commit 42f8a3aae55c21db433d6882bdf1994b2b7ae6f9 does not have https://github.com/ceph/ceph/pull/9374 merged.

gregsfortytwo:
...whoops, sorry. I forgot which pieces I had in which test branches. *blush*

gregsfortytwo:
On the down side, the reason I was running the test branch that way was to try and identify the cause of failing tests. At this point I can say conclusively that this PR is somehow causing us to leak inode references. :(

PR #1051, opened 2016-06-10
reference the kill_mds_by_rank function in the right location

gregsfortytwo:


jcsp:
Testing with https://github.com/ceph/ceph-qa-suite/pull/1053 here:
http://pulpito.ceph.com/jspray-2016-06-13_04:19:47-fs:thrash-master---basic-mira/

jcsp:
Passed here: http://pulpito.ceph.com/jspray-2016-06-13_06:12:14-fs:thrash-master---basic-mira/

PR #1052, opened 2016-06-13
rados/objectstore: pass CEPH_LIB env var to ceph_test_objectstore

tchaikov:
as the compressor plugins are installed into /usr/lib/ceph/compressor,
which is different from the building host. and ceph_test_objectstore's
default CEPH_LIB is $PWD/.libs. so we need to point
ceph_test_objectstore to the right place.


tchaikov:
fixes http://pulpito.ceph.com/kchai-2016-06-12_09:44:10-rados-wip-kefu-testing---basic-mira/254583/

tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-06-12_21:45:29-rados-master---basic-mira/

tchaikov:
fixes http://tracker.ceph.com/issues/16254

tchaikov:
superseded by https://github.com/ceph/ceph/pull/9662.

PR #1053, opened 2016-06-13
task/mds_thrash: don't drop exceptions from thrasher

jcsp:
Previously, when there was an exception in the thrasher
main loop, it would log it but ultimately "succeed"
the teuthology task.


jcsp:
Passed here (with #1051): http://pulpito.ceph.com/jspray-2016-06-13_06:12:14-fs:thrash-master---basic-mira/

jcsp:
@gregsfortytwo quick review?  This is the follow up to the issue you noticed

gregsfortytwo:
LGTM!

PR #1054, opened 2016-06-14
suites/fs: use simple messenger some places

jcsp:
Two tests are sensitive to the switch to async msgr
http://tracker.ceph.com/issues/16288
http://tracker.ceph.com/issues/16276


jcsp:
Tested here: http://pulpito.ceph.com/jspray-2016-06-14_09:32:43-fs:recovery-master---basic-mira/

PR #1055, opened 2016-06-15
16334 and 16332 fixes

yuriw:
Fixes http://tracker.ceph.com/issues/16334
Fixes http://tracker.ceph.com/issues/16332

### DIFF
--- a/suites/rados/singleton-nomsgr/all/lfn-upgrade-infernalis.yaml
+++ b/suites/rados/singleton-nomsgr/all/lfn-upgrade-infernalis.yaml
@@ -21,6 +21,7 @@ overrides:
     - scrub mismatch
     - ScrubResult
     - failed to encode
+    - wrongly marked me down
 roles:
 - - mon.a
   - osd.0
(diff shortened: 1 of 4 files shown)


suites/rados/singleton-nomsgr/all/lfn-upgrade-infernalis.yaml
```diff
@@ -21,6 +21,7 @@ overrides:
     - scrub mismatch
     - ScrubResult
     - failed to encode
+    - wrongly marked me down
```
athanatos: Need to get lfn-upgrade-hammer.yaml as well.

smithfarm:
@yuriw This happened in http://pulpito.ceph.com/smithfarm-2016-07-02_01:06:27-rados-hammer-backports---basic-smithi/287997 - which is `rados/singleton/{all/osd-backfill.yaml fs/btrfs.yaml msgr-failures/many.yaml}` so not exactly the same test.

PR #1057, opened 2016-06-20
tasks/scrub_test: use fuse path (do not assume filestore!)

liewegas:


liewegas:
http://pulpito.ceph.com/sage-scrub-a/267490/

PR #1058, opened 2016-06-21
tasks/ceph-deploy: Run on multiple Python versions (2, 3)

BlaXpirit:
This is tightly related to https://github.com/ceph/ceph-deploy/pull/388 (Python 3 Compatibility of ceph-deploy).
- The changes to the QA suite are needed in order to test ceph-deploy under different Python versions (here only major versions are specified - 2 or 3). If the *python_version* is not specified, nothing will be installed and the default *python* executable will be used (basically old behavior).
- And in turn, this pull request relies on the change of the bootstrap script in ceph-deploy, and, of course, the Python 3 compatibility is the new feature that is tested here.

Here are some sepia test runs under these changes:

- http://pulpito.ceph.com/oprypin-2016-06-16_08:41:28-ceph-deploy-jewel---basic-mira/
- http://pulpito.ceph.com/oprypin-2016-06-17_10:12:39-ceph-deploy-jewel---basic-mira/
- http://pulpito.ceph.com/oprypin-2016-06-20_15:51:37-ceph-deploy-jewel---basic-mira/
- http://pulpito.ceph.com/oprypin-2016-06-21_13:09:06-ceph-deploy-jewel---basic-mira/ 

There are failures but they seem to be caused by something else (such failures were known to happen before)
http://tracker.ceph.com/issues/16405

### DIFF
--- a/tasks/ceph_deploy.py
+++ b/tasks/ceph_deploy.py
@@ -24,7 +24,8 @@ def download_ceph_deploy(ctx, config):
     """
     Downloads ceph-deploy from the ceph.com git mirror and (by default)
     switches to the master branch. If the `ceph-deploy-branch` is specified, it
-    will use that instead.
+    will use that instead. The `bootstrap` script is ran, with the argument
+    obtained from `python_version`, if specified.
     """
     log.info('Downloading ceph-deploy...')
     testdir = teuthology.get_testdir(ctx)
@@ -38,14 +39,17 @@ def download_ceph_deploy(ctx, config):
             '{tdir}/ceph-deploy'.format(tdir=testdir),
         ],
     )
-    ceph_admin.run(
-        args=[
-            'cd',
-            '{tdir}/ceph-deploy'.format(tdir=testdir),
-            run.Raw('&&'),
-            './bootstrap',
-        ],
-    )
+    args = [
+        'cd',
+        '{tdir}/ceph-deploy'.format(tdir=testdir),
+        run.Raw('&&'),
+        './bootstrap',
+    ]
+    try:
+        args.append(str(config['python_version']))
+    except KeyError:
+        pass
+    ceph_admin.run(args=args)
 
     try:
         yield
--- a/tasks/ceph_deploy.py
+++ b/tasks/ceph_deploy.py
@@ -27,9 +27,38 @@ def download_ceph_deploy(ctx, config):
     will use that instead. The `bootstrap` script is ran, with the argument
     obtained from `python_version`, if specified.
     """
+    ceph_admin = ctx.cluster.only(teuthology.get_first_mon(ctx, config))
+
+    try:
+        py_ver = str(config['python_version'])
+    except KeyError:
+        pass
+    else:
+        supported_versions = ['2', '3']
+        if py_ver not in supported_versions:
+            raise ValueError("python_version must be: {}, not {}".format(
+                ' or '.join(supported_versions), py_ver
+            ))
+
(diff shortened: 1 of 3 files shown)


vasukulkarni:
@BlaXpirit  There are 2 tests that repeatedly failed with below error "Failed to create osd's", This I believe is due to python 3 changes https://paste.fedoraproject.org/382829/54919514/raw/

tasks/ceph_deploy.py
```diff
@@ -24,11 +24,47 @@ def download_ceph_deploy(ctx, config):
...
+            ))
+
+        log.info("Installing Python")
+        for admin in ceph_admin.remotes:
+            system_type = teuthology.get_system_type(admin)
+
+        if system_type == 'rpm':
+            if py_ver == '3':
+                ctx.cluster.run(args=[
+                    'sudo', 'yum', '-y', 'install', 'epel-release',
```
vasukulkarni: I dont think it epel-release should be required? epel is by default enabled on sepia labs.

tasks/ceph_deploy.py
```diff
@@ -24,11 +24,47 @@ def download_ceph_deploy(ctx, config):
...
+                    run.Raw('&&'),
+                    'sudo', 'yum', '-y', 'install',
+                    'python-virtualenv', 'python34'
```
vasukulkarni: python3 should be uninstalled if installed at the end of the test in finally: block

BlaXpirit:
@vasukulkarni That's the thing... if you examine the test runs, there isn't a single test that consistently fails every time, and @jdurgin says such errors were known to happen before

tasks/ceph_deploy.py
BlaXpirit: Why is that important?

Also... this has already been run on sepia so the hosts have python34 left on them

vasukulkarni:
@BlaXpirit I dont remember seeing that error before, here is the last run http://pulpito.ceph.com/teuthology-2016-06-21_02:55:02-ceph-deploy-jewel-distro-basic-vps/ ,  selinux denial issue is bit common which was seen in few of your runs.

tasks/ceph_deploy.py
vasukulkarni: The node is reused for other tests, so making python3 default on the node might actually break things which aren't ready for python3 yet and generally the test cleanup handles any things that it changed.

BlaXpirit: The default is not changed by installing another Python version. `python` always refers to Python 2.x, this just makes a `python3` available.

tasks/ceph_deploy.py
BlaXpirit: That seems to be the case. I will remove that if you think it's better.

jdurgin:
I left a comment on the tracker ticket pointing to a couple failures with jewel ceph-deploy

tasks/ceph_deploy.py
BlaXpirit: How can I check whether Python 3 was installed or it was already there?
We have to be absolutely sure, because Python 3 is an integral part of some distros.

jdurgin: it doesn't seem necessary to remove it to me

vasukulkarni:
@jdurgin thanks those look like ceph-disk issue that happens intermittently even on python2, I think maybe zap is holding the device long enough for the device in use issue, will check with loic.

vasukulkarni:
LGTM

PR #1059, opened 2016-06-22
Allow force-branch for s3roundtrip tests.

wusui:
Fixes tracker bug 15453
Signed-off-by: Warren Usui <wusui@redhat.com>

### DIFF
--- a/tasks/s3roundtrip.py
+++ b/tasks/s3roundtrip.py
@@ -28,13 +28,17 @@ def download(ctx, config):
     The context passed in should be identical to the context
     passed in to the main task.
     """
-    assert isinstance(config, list)
+    assert isinstance(config, dict)
     log.info('Downloading s3-tests...')
     testdir = teuthology.get_testdir(ctx)
-    for client in config:
+    for (client, cconf) in config.iteritems():
+        branch = cconf.get('force-branch', None)
+        if not branch:
+            branch = cconf.get('branch', 'master')
         ctx.cluster.only(client).run(
             args=[
                 'git', 'clone',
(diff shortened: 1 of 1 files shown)


tasks/s3roundtrip.py
```diff
@@ -28,13 +28,17 @@ def download(ctx, config):
     The context passed in should be identical to the context
     passed in to the main task.
     """
-    assert isinstance(config, list)
+    assert isinstance(config, dict)
```
zmc: Will this break existing tests?

zmc:
Looks good aside from the one concern

yuriw:
@wusui we use `force-branch` in suites, upgrades, pls test

tasks/s3roundtrip.py
wusui: I looked a little further and I probably should make a change to the stuff in the finally: section of this routine.

wusui: Upon further review, I believe that the finally: section is correct.  I have tested this with force-branch ceph-master, force-branch aardvark, and force-branch hammer and all three cases seem to behave as expect (aardvark is an error).

zmc: I was actually referring specifically to line 31 there :)

wusui: Yes. I changed that because I changed the downloads call to be a dict so that the new force-branch parameter can be used.  The only uses of the config parameter inside this function is in the code that I modified and in the finally code that I just checked.  I was afraid that this may break things but it looks to be correct.

PR #1064, opened 2016-06-24
tasks/cephfs: vstart_runner fixes

jcsp:
Mixture of things that tripped up while running tests all together on a vstart cluster

### DIFF
--- a/tasks/cephfs/test_cap_flush.py
+++ b/tasks/cephfs/test_cap_flush.py
@@ -4,9 +4,10 @@ import time
 from textwrap import dedent
 from unittest import SkipTest
 from tasks.cephfs.fuse_mount import FuseMount
-from tasks.cephfs.cephfs_test_case import CephFSTestCase
+from tasks.cephfs.cephfs_test_case import CephFSTestCase, long_running
 
 class TestCapFlush(CephFSTestCase):
+    @long_running
     def test_replay_create(self):
         """
         MDS starts to handle client caps when it enters clientreplay stage.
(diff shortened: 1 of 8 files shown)


tasks/cephfs/test_pool_perm.py
```diff
@@ -80,6 +80,7 @@ def test_forbidden_modification(self):
 
         self.mount_a.umount_wait()
         self.mount_a.mount()
+        self.mount_a.wait_until_mounted()
```
ajarr: So I guess we can be more careful and start using the 'mounted' context manager more widely across the code base and not worry too much about races and if we forgot to umount?
So just,
with self.mount_a.mounted():
as done in 'test_journal_migration.py'

tasks/cephfs/test_cap_flush.py
```diff
@@ -4,9 +4,10 @@
 from textwrap import dedent
 from unittest import SkipTest
 from tasks.cephfs.fuse_mount import FuseMount
-from tasks.cephfs.cephfs_test_case import CephFSTestCase
+from tasks.cephfs.cephfs_test_case import CephFSTestCase, long_running
```
ajarr: do you think I can change the long_running to something like {with, for, use}_teuthology as a follow up patch?

tasks/cephfs/test_pool_perm.py
jcsp: Yes, we should make more use of that

ajarr:
Code looks good to me.

PR #1065, opened 2016-06-24
rados/thrash/workloads/cache-agent-big: set min_size=2 on m=1 k=2 ec pool

liewegas:
Otherwise min_size will be 3 and we will wedge on a single down
osd.


dachary:
LGTM

PR #1066, opened 2016-06-28
cephfs: add test for dump tree admin socket command

fullerdj:
Add test_dump_tree to validate the MDS admin socket command
dump tree. Add a parameter for path_to_ino to enable the use of
lstat() instead of stat()

See: http://tracker.ceph.com/issues/11171
See: https://github.com/ceph/ceph/pull/9925
Signed-off-by: Douglas Fuller <dfuller@redhat.com>

### DIFF
--- a/suites/fs/basic/tasks/asok_dump_tree.yaml
+++ b/suites/fs/basic/tasks/asok_dump_tree.yaml
@@ -1,0 +1,7 @@
+tasks:
+- install:
+- ceph:
+- ceph-fuse:
+- cephfs_test_runner:
+    modules:
+      - tasks.cephfs.test_dump_tree
--- a/tasks/cephfs/test_dump_tree.py
+++ b/tasks/cephfs/test_dump_tree.py
@@ -1,0 +1,66 @@
+from tasks.cephfs.cephfs_test_case import CephFSTestCase
+import random
+import os
+
+class TestDumpTree(CephFSTestCase):
+    def get_paths_to_ino(self):
+        inos = {}
+        p = self.mount_a.run_shell(["find", "./"])
+        paths = p.stdout.getvalue().strip().split()
+        for path in paths:
+            inos[path] = self.mount_a.path_to_ino(path, False)
+
+        return inos
+
+    def populate(self):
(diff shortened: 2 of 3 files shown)


tasks/cephfs/test_dump_tree.py
```diff
@@ -0,0 +1,46 @@
...
+        for path in paths:
+            inos[path] = self.mount_a.path_to_ino(path, False)
+
+        return inos
+
+    def populate(self):
+        self.mount_a.run_shell(["git", "clone",
+                                "https://github.com/ceph/ceph-qa-suite"])
+        
+    def test_dump_tree(self):
```
jcsp: Hmm, why do we need a lot of files and then to pick random ones from within that?  Putting randomness into tests feels like it might be asking for trouble.

What if you created these paths:
```
alpha1/
alpha1/bravo
alpha1/bravo/charlie
alpha2/
```
Then test that dump on / includes on everything, dump on alpha1/ with depth high excludes alpha2, and dump on alpha1 with depth=1 excludes charlie.  Would that be missing any cases?

jcsp:
Looks good, thanks for updating.  Will include in next test run + then merge.

suites/fs/basic/tasks/asok_dump_tree.yaml
```diff
@@ -0,0 +1,4 @@
+tasks:
+- cephfs_test_runner:
```
jcsp: In fs/basic the mounts are defined in the tasks (like in the cfuse_*) ones.  So this is failing to run because it doesn't have any mounts.  (Different from the ones in fs/recovery which have the mounts defined elsewhere fs/recovery/mounts)

suites/fs/basic/tasks/asok_dump_tree.yaml
```diff
@@ -0,0 +1,7 @@
+tasks:
+- install:
```
jcsp: You don't need the install and ceph bits here (they're in the inline/) tasks, perversely...

jcsp:
Merged manually with an extra commit to fix the yaml

PR #1069, opened 2016-06-29
cephfs: test fragment size limit

batrick:
The fragment configuration uses 10000 for the fragment max size. The reason for
this is that many tests add 1000 files to a single directory which will hit
this limit without fragmentation catching up.

The test_dirfrag_limit test confirms:

o That the directory fragment size cannot exceed mds_bal_fragment_size_max (using a limit of 50 in all configurations).
o That fragmentation (forced) will allow more entries to be created.
o That unlink fails when the stray directory fragment becomes too large and that unlinking may continue once those strays are purged.

Tests: https://github.com/ceph/ceph/pull/9789
Issue: http://tracker.ceph.com/issues/16164

### DIFF
--- a/suites/fs/basic/dirfrag/frag_enable.yaml
+++ b/suites/fs/basic/dirfrag/frag_enable.yaml
@@ -4,6 +4,7 @@ overrides:
     conf:
       mds:
         mds bal frag: true
+        mds bal fragment size max: 10000
         mds bal split size: 100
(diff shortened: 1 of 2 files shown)


batrick:
I've already run this through teuthology. There were two failures (valgrind warning on the client segfault in the client) which were unrelated to this test. I'm currently looking into the segfault.

PR #1073, opened 2016-07-01
fs: add snapshot tests to mds thrashing

ukernel:


gregsfortytwo:
Is there anything left to stop us merging this @jcsp, @ukernel? I presume the test run from https://github.com/ceph/ceph/pull/9955 included this so we believe it passes?

PR #1077, opened 2016-07-04
rbd: tests with rbd_skip_partial_discard option enabled

trociny:
Signed-off-by: Mykola Golub <mgolub@mirantis.com>


dillaman:
Would you mind combining the partial discard and copy-on-read config options into a new "config" directory where we just test no overrides, copy-on-read, and skip partial discard?

dillaman:
@trociny

trociny:
@dillaman sorry, I missed your note due to wrong mail filter.

Do you mean something like the new patch? I will squash it later.

dillaman:
@trociny Perfect -- thx

trociny:
@dillaman rebased

PR #1079, opened 2016-07-08
Ensure the ceph-deploy tests work on smithi with nvme partitions

vasukulkarni:
skip zap disk and use activate for nvme partitions


liewegas:
hmm, this seems like a hack that assumes pre-partitioned nvme.. i wonder if there is a better way to get this behavior.  

although i'm not sure how important it is that we test ceph-deploy on pre-partitioned drives anyway?

vasukulkarni:
closing this for now, as it gets covered on other than smithi nodes.

PR #1080, opened 2016-07-11
tasks/cephfs/test_volume_client: test authentication metadata

ajarr:
tasks/cephfs/test_volume_client: test authentication metadata

Test the use cases for the authentication metadata stored
by the volume client:

* Obtain the list of auth IDs having access to a volume.

* Restrict volume access to auth IDs of a single (OpenStack)
  tenant to enforce strong tenant isolation of volumes.


ajarr:
partially tests https://github.com/ceph/ceph/pull/9864

PR #1082, opened 2016-07-13
tasks/cephfs: check cache size health warning

jcsp:
Fortunately we already have a test that creates the condition,
so just tweak it to exceed the 150% threshold for the health warning,
and check that the health message appears.


jcsp:
Tested here http://pulpito.ceph.com/jspray-2016-07-11_03:45:31-fs-wip-mds-minor-distro-basic-mira/

jcsp:
Oops, this should have merged with the associated ceph change

PR #1084, opened 2016-07-15
task/packages: install libcephfs1-dev

jcsp:
The way we currently build the cephfs java
bindings depends on the unversioned .so file
which is in the -dev package, not the
main one.

Fixes: http://tracker.ceph.com/issues/16640


jcsp:
Tested here: http://qa-proxy.ceph.com/teuthology/jspray-2016-07-29_06:53:13-fs:basic-master-distro-basic-mira/339468/teuthology.log

PR #1087, opened 2016-07-20
rados: Added tests for the striper layer of rados

sponce:
Signed-off-by: Sebastien Ponce sebastien.ponce@cern.ch


tchaikov:
pushed to wip-1087 for testing.

PR #1091, opened 2016-07-22
11429: wait_for_clean between restarting the osds and starting the bench

athanatos:
ceph.restart now marks the osds down, so the objects are actually being
created while slowest of the osds boots. That causes a ton of 1 byte
objects to be created in a degraded state and causes the cleanup to take
a long time.  Also, reduce length of bench since it's only being used
to ensure the osds came up correctly.


smithfarm:
@athanatos Does it make sense to cherry-pick this to jewel and hammer as well?

athanatos:
@smithfarm Yeah

PR #1094, opened 2016-07-25
Reduce number of jobs for hammer branch

yuriw:


smithfarm:
@yuriw: Could you add an elif for jewel that would use `/40` ? That would give 348 jobs for jewel rados runs.

PR #1097, opened 2016-07-25
Wip 16413 hammer

smithfarm:


athanatos:
Some of those do not make sense.  bc4e461 and 	67c6828 are fixes for stopping first at hammer before upgrading to things after hammer.  That doesn't make sense to backport.  I *think* the others make sense, but you need to actually run the suite to make sure they aren't going to break anything.

smithfarm:
The other commits do not cherry-pick cleanly without the first two. I'll focus first on getting the test to run at all in hammer, and then look at optimizing it.

PR #1108, opened 2016-08-05
suites/fs: remove "32bit"

jcsp:
Presumably at some point this was being scheduled on 32 bit builds,
but I don't think it is now, and we're withdrawing 32 bit ceph.  Saves
us two jobs in our fs suite runs.


gregsfortytwo:
Do we really want to kill the 32-bit clients? :/ (That's what this test is for.)

If we do we should probably rip out all the fake ino code at the same time, but I'd really rather we keep testing it given the push on ARM builds and stuff.

jcsp:
Oh, never mind, I thought this was cruft because it wasn't being run against any 32 bit builds, but now I see that it's really the "use_faked_inos" test as opposed to actually running on anything 32 bit.

PR #1109, opened 2016-08-08
fs: fuse_default_permissions = 0 for kernel build test

batrick:
Ported from Zheng's #22: 2e283ce6d77c66a564d807f0b0909b238c4fb5f0

"This can reduce the test time becuase it avoids sending getattr request
whenever the kernel checks inode permission."

This is part of an effort to eliminate unnecessary differences between
multimds and fs suites.

### DIFF
--- a/suites/fs/basic/tasks/cfuse_workunit_kernel_untar_build.yaml
+++ b/suites/fs/basic/tasks/cfuse_workunit_kernel_untar_build.yaml
@@ -1,3 +1,8 @@
+overrides:
+  ceph:
+    conf:
+      client:
+        fuse_default_permissions: 0
 tasks:
 - ceph-fuse:
 - workunit:


batrick:
@ukernel, comments?

batrick:
I'm actually confused by this original PR (#22) now. Isn't a false value going to **turn on** the `getattr`s?

https://github.com/ceph/ceph/blob/master/src/client/Client.cc#L9520-L9524

gregsfortytwo:
We're *disabling* FUSE's own permission-checking code and turning on the Client's:
https://github.com/ceph/ceph/blob/master/src/client/fuse_ll.cc#L995-L998

Our internal code is a lot more efficient about only grabbing what pieces we need than FUSE's, which does constant getattrs() in order to do the permission checking.

batrick:
Ah, got it. Thanks @gregsfortytwo!

PR #1110, opened 2016-08-08
cephfs/vstart_runner: add -w option to ps

fullerdj:
vstart_runner can't find arguments to ceph daemons to identify them with
ps -x because commands are cut off at terminal width. Add -w for wide
output.

Signed-off-by: Douglas Fuller <dfuller@redhat.com>

### DIFF
--- a/tasks/cephfs/vstart_runner.py
+++ b/tasks/cephfs/vstart_runner.py
@@ -257,7 +257,7 @@ class LocalDaemon(object):
         Return PID as an integer or None if not found
         """
         ps_txt = self.controller.run(
-            args=["ps", "-xu"+str(os.getuid())]
+            args=["ps", "-xwwu"+str(os.getuid())]
         ).stdout.getvalue().strip()
         lines = ps_txt.split("\n")[1:]
 


gregsfortytwo:
We've already done something like this once -- was that for teuthology and we missed it in vstart? Or do we have an ongoing problem?

jcsp:
Deja vu from this one:
<pre>
commit 8e6a164dfb957a838482207a911289ffab0936f2
Author: John Spray <john.spray@redhat.com>
Date:   Thu May 12 11:16:56 2016 +0100

    tasks/cephfs: fix ps usage to not truncate cmds
    
    Now that daemon commandlines include "--cluster" they
    easily overflow ps's default behaviour.
    
    Fixes: http://tracker.ceph.com/issues/15857

</pre>

tasks/cephfs/vstart_runner.py
```diff
@@ -257,7 +257,7 @@ def _get_pid(self):
         Return PID as an integer or None if not found
         """
         ps_txt = self.controller.run(
-            args=["ps", "-xu"+str(os.getuid())]
+            args=["ps", "-xwu"+str(os.getuid())]
```
jcsp: Let's use ``ww`` instead of ``w`` (make it unlimited instead of exchanging one limit for another)

jcsp:
Looks sane, see comment about ww

fullerdj:
Good point. Fixed.

PR #1113, opened 2016-08-09
cephfs: Test recovery into an alternate pool

fullerdj:
Parameterize test_data_scan to support testing recovery into an alternate
metadata pool. Add one test case to exercise this path.

cf. https://github.com/fullerdj/ceph/tree/wip-djf-15069

Signed-off-by: Douglas Fuller dfuller@redhat.com


tasks/cephfs/fuse_mount.py
```diff
@@ -60,6 +64,9 @@ def mount(self, mount_path=None):
             self.mountpoint,
         ]
 
+        if client_mds_namespace:
```
jcsp: should be self.client_mds_namespace

fullerdj:
Fixed, thanks.

liewegas:

PR #1118, opened 2016-08-14
tasks/scrub_test: use fuse path (do not assume filestore!)

liewegas:


tchaikov:
@liewegas this change was merged before as #1057 and then reverted by #1063. as it assumes the layout of the files is set up by FuseStore, the scrub_test will fail with FileStore i believe. so shall i close this PR? unless we have a better way to make this test more flexible.

liewegas:
@tchaikov FuseStore works with FileStore too.  e.g., http://pulpito.ceph.com/sage-scrub-c/411754

liewegas:
https://github.com/ceph/ceph/pull/11057

PR #1122, opened 2016-08-17
Fix big logs in rados thrash runs

dzafman:


dzafman:
Testing shows that the fix (1 commit) for admin socket objecter requests test works:

http://pulpito.ceph.com/dzafman-2016-08-10_18:47:06-rados:thrash-master-distro-basic-smithi

dzafman:
10 rados:thrash runs show reasonable log sizes.

http://pulpito.ceph.com/dzafman-2016-08-17_16:18:20-rados:thrash-wip-zafman-testing-distro-basic-smithi/

dzafman:
Now final fixes squashed.

tasks/ceph_manager.py
```diff
@@ -727,11 +731,11 @@ def do_dump_ops(self):
             for osd in self.live_osds:
                 # Ignore errors because live_osds is in flux
                 self.ceph_manager.osd_admin_socket(osd, command=['dump_ops_in_flight'],
-                                     check_status=False, timeout=30)
+                                     check_status=False, timeout=30, stdout=DEVNULL)
```
tchaikov: if we don't want `do_dump_ops()`, why not just disable it using the  `dump_ops_enable` option?

and the whole point of `do_dump_ops()` is for dumping the ops, if we simply redirect the stdout to `/dev/null`, we render this method useless.

dzafman: It is a thasher test to see if the OSD crashes or if there is no locking deadlock or something.  No code is examining the output for correctness.

dzafman:
I'm testing to see if I really need the admin_socket test to disable dump_ops thrashing.  If not I'll delete that commit.

dzafman:
Passes http://pulpito.ceph.com/dzafman-2016-08-18_21:12:37-rados:thrash-wip-zafman-testing-distro-basic-smithi/

tasks/ceph_manager.py
tchaikov: if we don't want the side effects of `do_dump_ops()`, probably we can just remove this method? i.e. partially revert 7a528763d126eaca90e40b02aad02ab9c3c92a32 ?

PR #1124, opened 2016-08-18

jdurgin:

This requires https://github.com/ceph/ceph/pull/10776


tchaikov:
lgtm.

PR #1126, opened 2016-08-19
Add the option to run rados tests on Python 3

oprypin:
This adds some test items to the rados suite that run Python 3 instead of Python 2.

A few notes:
- https://github.com/ceph/ceph/pull/10782 allows specifying the Python executable for _rados/test_python.sh_ as an environment variable
- https://github.com/ceph/teuthology/pull/922 allows installing the packages as just _python3-*_ although they're actually _python34-*_ on CentOS.
- The recurring `sed` usage replaces the hashbang of the _ceph_ script to contain `#!/usr/bin/python3`.

### DIFF
--- a/suites/rados/basic/tasks/rados_python3.yaml
+++ b/suites/rados/basic/tasks/rados_python3.yaml
@@ -1,0 +1,15 @@
+tasks:
+- install:
+    extra_packages: [python3-rados]
+- exec:
+    all:
+      - 'sudo sed -i -e "1s,^\(#!/usr/bin.*python\)[0-9]*$,\13," /usr/bin/ceph'
+- ceph:
+    log-whitelist:
+    - wrongly marked me down
+- workunit:
+    env:
+      PYTHON: python3
+    clients:
+      client.0:
+        - rados/test_python.sh
--- a/suites/rados/singleton/all/cephtool-python3.yaml
+++ b/suites/rados/singleton/all/cephtool-python3.yaml
@@ -12,9 +12,10 @@ openstack:
       size: 10 # GB
 tasks:
 - install:
+    extra_packages: [python3-cephfs, python3-rados, python3-rbd]
 - exec:
     all:
-      - 'sudo sed -i -e "1s,^\(#!/usr/bin.*python\)[0-9]*$,\12," /usr/bin/ceph'
+      - 'sudo sed -i -e "1s,^\(#!/usr/bin.*python\)[0-9]*$,\13," /usr/bin/ceph'
 - ceph:
     log-whitelist:
     - wrongly marked me down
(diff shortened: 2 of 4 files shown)


BlaXpirit:
Some test runs of this:
- http://pulpito.front.sepia.ceph.com/oprypin-2016-08-18_21:54:21-rados-master---basic-mira/
- http://pulpito.front.sepia.ceph.com/oprypin-2016-08-18_21:43:22-rados-master---basic-mira/

jdurgin:
lgtm. @athanatos any objection to adding this as another facet of the rados:thrash suite?

suites/rados/basic/tasks/rados_python3.yaml
```diff
@@ -0,0 +1,15 @@
+tasks:
+- install:
+    extra_packages: [python3-rados]
+- exec:
+    all:
+      - 'sudo sed -i -e "1s,^\(#!/usr/bin.*python\)[0-9]*$,\13," /usr/bin/ceph'
```
athanatos: Is there a better way to do this?

athanatos:
I think this is a good idea, but is using sed to modify the ceph script the really best way to do it?

BlaXpirit:
It's the easiest and most reliable way that I can think of. The alternative would be to scan the whole codebase for possible users of the ceph executable and explicitly call it through some version of Python, passed through some variable.

Of course, this idea may be approached from a different angle, like exposing a generic hook to perform this action in Python code instead of repeating it in yaml...

athanatos:
Is this even correct for non-ubuntu?

athanatos:
In ceph.in I see @PYTHON_EXECUTABLE@ which suggests to me that we are doing platform specific path wrangling in ceph.py.  Why don't we use /usr/bin/env?

athanatos:
Oh, well, looks like this is the only way to do it.  Bad python, no cookie.  lgtm

BlaXpirit:
I don't see how it's not correct on non-Ubuntu. `python`, `python2`, `python3` all work as expected on CentOS 7, with or without `env`.

@PYTHON_EXECUTABLE@ just ends up being `/usr/bin/python`, it's currently Python 2 specifically in the CMake builds.

athanatos:
This is ok with me, but @ktdreyer suggested adding a package which adds a /usr/bin/ceph-python3 (or something) file to force python3.  That seems like a better approach to me (though it would be more work).

If you decide to go with this verison, I want you to run at least 3 thrasher tests with the python3 modification above.  Your test runs above don't seem to run anything which exercises the thrasher variants, and those are the ones I'd be most worried about breakage in.

liewegas:

PR #1127, opened 2016-08-19
add test case for envlibrados

ryneli:


jdurgin:
this looks good, will merge once the workunit is merged

PR #1128, opened 2016-08-22
ceph_manager: lower mon_reweight_min_bytes_per_osd to 10

tchaikov:
otherwise monitor could reject the command:
```
Refusing to reweight: we only used 588084 kb used across all osds!
```
if the average used space is smaller than
`mon_reweight_min_bytes_per_osd`.

Fixes: http://tracker.ceph.com/issues/16805


tchaikov:
tested at http://pulpito.front.sepia.ceph.com:80/kchai-2016-08-22_17:13:12-rados-master---basic-mira/

athanatos:
lgtm

PR #1130, opened 2016-08-22
Add firefly packages for install task

vasukulkarni:
Add firefly packages for install task


vasukulkarni:
Testing install task

vasukulkarni:
Works on ubuntu/centos http://pulpito.front.sepia.ceph.com/vasu-2016-08-22_22:32:02-smoke-firefly---basic-vps/

vasukulkarni:
@jdurgin @smithfarm  can one of you take a look at this and merge for firefly install task to work and for other suites which are dependent on this(mostly for some rare runs)

smithfarm:
@vasukulkarni Is this to address http://tracker.ceph.com/issues/17075 which was caused by https://github.com/ceph/teuthology/commit/4543fb64 ?

smithfarm:
@zmc @vasukulkarni I originally thought we could drop the python-ceph package in, say, kraken and deal with that by overriding the package list for individual tests. Overrides would ensure that the tests that install pre-hammer versions have the right package list (i.e. python-ceph instead of the three packages that replace it) *and* they would have it regardless of which suite-branch is given. (This PR only works if `--suite-branch firefly` is given when scheduling the test.)

Later, though, I realized that the package list is not used just at the beginning of the test, but also at the end - in the cleanup phase. For tests that install firefly and then upgrade, I'm afraid there's no way to provide a correct package list, i.e. one that will work at the beginning (for firefly installation) *and* at the end (for post-firefly removal).

vasukulkarni:
closing since the old package list has been restored

PR #1135, opened 2016-08-23
ceph_manager: do not use StringIO instance as default argument

athanatos:
Otherwise, all invokations share the *same* StringIO instance
leading to some very silly bugs.


jtlayton:
ACK from my standpoint. That seems to fix the fs suite test failures we have been seeing for the last week or so.

zmc:
👍

PR #1143, opened 2016-08-25
workunit: allow parallel git clone

smithfarm:
When multiple client run in parallel on the same machine and they
try to get workunits from a repository that is not github, they must
git clone in a directory that is suffixed as srcdir. Otherwise they
will conflict with each other.

Fixes: http://tracker.ceph.com/issues/17116

(cherry picked from commit b9813e490ec8a5bb59e177612f047899397c342e)


smithfarm:
tested in #1134

PR #1145, opened 2016-08-26
mds: add tests for mantle (programmable balancer)

michaelsevilla:
Tests load/execute balancers but doesn't test balancing behavior; should be ok as long as Mantle is only used by expert experimenters.

Signed-off-by: Michael Sevilla <mikesevilla3@gmail.com>


jcsp:
ok to test

jcsp:
add to whitelist

PR #1149, opened 2016-08-30
krbd/unmap: assorted pre-single-major fixes

idryomov:


idryomov:
The fix for the kernel task regression is at https://github.com/ceph/teuthology/pull/945.

PR #1153, opened 2016-08-31
tasks/ceph: construct CephManager earlier

jcsp:
Previously, if errors occurred during healthy(), then
the finally block would invoke osd_scrub_pgs, which relies
on CephManager being constructed, and it would die, hiding
the original exception.

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -1485,19 +1485,21 @@ def task(ctx, config):
     ]
 
     with contextutil.nested(*subtasks):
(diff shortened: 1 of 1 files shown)


jcsp:
Used here without issues http://pulpito.ceph.com/jspray-2016-09-05_10:01:52-fs-wip-jcsp-testing-20160902-testing-basic-mira/

batrick:
+1

zmc:
:thumbsup:

vasukulkarni:
I believe a pg scrub would be not so useful after this task failure, but nice to exit sooner than later.

PR #1141, opened 2016-08-25
buildpackages: make make-{deb,rpm}.sh aware of cmake

rjfd:
This is the followup of an existing PR #1131 

Fixes: http://tracker.ceph.com/issues/16819

Signed-off-by: Ricardo Dias rdias@suse.com
Signed-off-by: Nathan Cutler ncutler@suse.com

PR #1162, opened 2016-09-05
buildpackages/make-rpm.sh: use /etc/os-release

smithfarm:
The lsb_release binary is deprecated and requires installation of packages.

The /etc/os-release file is guaranteed to be present.

Signed-off-by: Nathan Cutler ncutler@suse.com

### DIFF
--- a/tasks/buildpackages/make-rpm.sh
+++ b/tasks/buildpackages/make-rpm.sh
@@ -43,10 +43,9 @@ source $(dirname $0)/common.sh
 
 init_ceph $git_ceph_url $sha1
 
-#id=$(lsb_release -s -i | tr A-Z a-z)
-#major=$(lsb_release -s -r | sed -s "s;\..*;;g")
-#codename="${id}${major}"
-releasedir=$base/$(lsb_release -si | tr ' ' '_')/WORKDIR
+distro=$( source /etc/os-release ; echo $ID )
+distro_version=$( source /etc/os-release ; echo $VERSION )
+releasedir=$base/$distro/WORKDIR
 #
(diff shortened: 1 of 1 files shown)


smithfarm:
@rjfd

smithfarm:
Marking DNM until I manage to run a test.

smithfarm:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zTwwnoGR&#x2F;1973-buildpackages-make-rpm-sh-use-etc-os-release"> buildpackages&#x2F;make-rpm.sh: use &#x2F;etc&#x2F;os-release</a>

rjfd:
The new information extracted from `os-release` is different from the lsb_release binary, at least for OpenSUSE Leap (where I tested), but I think it does not break anything, it just gives you a different `releasedir` path.

Therefore it LGTM!

rjfd:
I remember to also see references to `lsb_release` inside teuthology code, maybe we should also make a similar change there.

smithfarm:
This does not work on its own. First of all, it needs https://github.com/ceph/ceph/pull/10981 to be merged, otherwise `install-deps.sh` fails. Second, I found that #1141 was not sufficient to fix `make-rpm.sh` for CentOS 7.2 so I'm working on another patch for that.

smithfarm:
CentOS:

**running** http://158.69.89.232:8081/ubuntu-2016-09-05_17:21:00-buildpackages:any-master---basic-openstack/

smithfarm:
@dachary This fixes `make-rpm.sh` on master . . . again.

PR #1165, opened 2016-09-06
rados/singleton-nomsgr/all/11429: wait until clean before checking log

tchaikov:
to make sure that load_pgs() is finished before checking its output

Fixes: http://tracker.ceph.com/issues/16157


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-09-06_08:57:14-rados-wip-kefu-testing---basic-mira/

PR #1166, opened 2016-09-06
[DNM] tasks/cephfs : test recovery of partial auth update

ajarr:
... in ceph_volume_client.

Signed-off-by: Ramana Raja <rraja@redhat.com>


jcsp:
I guess this is no longer DNM when combined with https://github.com/ceph/ceph/pull/11304?  Putting this in for testing along with that.

ajarr:
The test case was partial. I'll test post recovery case in a separate PR.

PR #1167, opened 2016-09-06
Backport krbd/unmap fixes

idryomov:
http://tracker.ceph.com/issues/17222


idryomov:
Test run: http://pulpito.ceph.com/dis-2016-09-06_18:07:15-krbd:unmap-jewel-testing-basic-mira/

PR #1168, opened 2016-09-07
Add writethrough flush tests

vasukulkarni:
librbd writethrough flush tests that can be compared using fio's bw/rw/latency/iops etc

### DIFF
--- a/suites/rbd/singleton/all/librbd-writethrough-flush-false.yaml
+++ b/suites/rbd/singleton/all/librbd-writethrough-flush-false.yaml
@@ -1,0 +1,19 @@
+roles:
+- [mon.a, osd.0, osd.1, osd.2]
+- [osd.3, osd.4, osd.5]
+- [client.0]
+tasks:
+- install:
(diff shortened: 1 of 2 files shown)


vasukulkarni:
test runs on smithi http://pulpito.ceph.com/vasu-2016-09-07_19:20:32-rbd:singleton-jewel---basic-smithi/

dillaman:
@vasukulkarni fio doesn't issue a flush request -- so these tests are effectively the same test. There is an effort to integrate CBT into teuthology to test for performance regressions. I think this is the sort of effort that is probably best directed there.

vasukulkarni:
@dillaman thanks, wondering if there is any internal tool or something to verify for functionality other than performance gain.

dillaman:
@vasukulkarni perhaps using qemu_bonnie tests would work until the CBT stuff is integrated

vasukulkarni:
will test it out with qemu_bonnie and open again

PR #1170, opened 2016-09-09
rbd: move nbd test workload to separate client host from OSDs

dillaman:


trociny:
lgtm

PR #1172, opened 2016-09-12
suites/upgrade: override rbd_default_features config back to pre-Jewel

trociny:
Fixes: http://tracker.ceph.com/issues/16717
Signed-off-by: Mykola Golub <mgolub@mirantis.com>


trociny:
@dillaman Sorry for the delay -- I have missed your note in the tracker.

trociny:
Here is an example of the successful run with the patch applied:

http://pulpito.ceph.com/trociny-2016-09-12_11:25:36-upgrade:client-upgrade-jewel---basic-vps/

For the reference here is a failed run without the patch:

http://pulpito.ceph.com/trociny-2016-09-12_11:26:06-upgrade:client-upgrade-jewel---basic-vps/

PR #1175, opened 2016-09-14
MDS thrasher for multimds

batrick:
```
This commit amends the MDS thrasher task to also work on multimds
clusters. Main changes:

o New FSStatus class in tasks/cephfs/filesystem.py which gets a snapshot
  of the fsmap (`ceph fs dump`). This allows consecutive operations on
  the same fsmap without repeated fs dumps.

o Only one MDSThrasher is started for each file system.

o The MDSThrasher operates on ranks instead of names (and groups of
  standbys following the initial active).

o The MDSThrasher also will change the max_mds for the cluster to a new
  value [1, current) or (current, starting max_mds]. When reduced,
  randomly selected MDSs other than rank 0 will be deactivated to reach
  the new max_mds. The likelihood of changing max_mds in a given cycle of
  the MDSThrasher is set by the "thrash_max_mds" config.

o The MDSThrasher prints out stats on completion, e.g. number of
  mds deactivated or mds_max changed.
```


batrick:
This is now based on master and ready for merging. Latest test run:

http://pulpito.ceph.com/pdonnell-2016-09-28_19:51:25-fs:thrash-master---basic-mira/

The dead job is due to inability lock enough mira machines. Only unusual job was:

http://pulpito.ceph.com/pdonnell-2016-09-28_19:51:25-fs:thrash-master---basic-mira/442586/

which ran for 8 hours. That doesn't appear to be normal but doesn't appear related to the thrasher:

````
2016-09-29T01:12:28.604 INFO:teuthology.task.pcp:PCP+Grafana dashboard: http://pcp.front.sepia.ceph.com:44323/grafana/index.html#/dashboard/script/index.js?time_to=2016-09-29T01%3A12%3A28&time_from=2016-09-29T00%3A58%3A27&hosts=mira038%2Cmira025
2016-09-29T09:07:52.954 DEBUG:teuthology.run_tasks:Unwinding manager internal.timer
````

batrick:
John, please take a look when you have time.

jcsp:
Several tests fail because get_fs_map function is gone

http://pulpito.ceph.com/jspray-2016-10-10_18:50:24-fs-master-distro-basic-mira

There's also a failure from a scrub message in log which will go away with a rebase or on merge, and a PJD failure which is scary but presumably nothing to do with this

batrick:
Okay this should be ready now. Last full fs test run:

http://pulpito.ceph.com/pdonnell-2016-11-07_16:09:36-fs-master---basic-mira/

The dead job looks like some kind of teuthology failure with an osd gone missing. The only failure was fixed in 9832bea and tested in:

http://pulpito.ceph.com/pdonnell-2016-11-08_02:37:39-fs-master---basic-mira/

jcsp:
I noticed that this broke vstart runner, please pull in these commits: https://github.com/batrick/ceph-qa-suite/pull/1

batrick:
@jcsp, merged!

PR #1176, opened 2016-09-19
tasks/cephfs: add default paths to vstart_runner

jcsp:
So that for folks with sources in typical locations
(or typical on my workstation at least!) invoking
vstart_runner is less of a mouthful.


jcsp:
@batrick would you like to review this?

jcsp:
@batrick updated with more paths, shall we merge?

PR #1180, opened 2016-09-21
misc fixes to test bluestore, but not actual tests in the suite (yet)

liewegas:


tasks/ceph_manager.py
```diff
@@ -1926,6 +1926,21 @@ def kill_osd(self, osd):
...
             remote.console.power_off()
+        elif self.config.get('bdev_inject_crash') and self.config.get('bdev_inject_crash_probability'):
+            if random.uniform(0, 1) < self.config.get('bdev_inject_crash_probability'):
+                self.raw_cluster_cmd(
+                    '--', 'tell', 'osd.%d' % osd,
+                    'injectargs',
+                    '--bdev-inject-crash %d' % self.config.get('bdev_inject_crash'),
+                )
+                try:
+                    self.ctx.daemons.get_daemon('osd', osd).wait()
```
jdurgin: this and the other get_daemon() call should pass self.cluster as a third parameter

tasks/thrashosds.py
```diff
@@ -159,8 +159,12 @@ def task(ctx, config):
                         'but not available on osd role: {r}'.format(
                             r=remote.name))
 
-    log.info('Beginning thrashosds...')
     cluster_manager = ctx.managers[cluster]
+    for f in ['powercycle', 'bdev_inject_crash']:
```
jdurgin: the bdev_inject_* options are missing from the task() docstring

PR #1186, opened 2016-09-23
suites/fs: pin java test to ubuntu

jcsp:
Workaround for #16640


jcsp:
(This works nicely http://pulpito.ceph.com/jspray-2016-09-23_00:10:55-fs-wip-jcsp-greenish-distro-basic-smithi/)

gregsfortytwo:
We're really boxing ourselves in with these ubuntu-only tests. :(
(All our samba ones are ubuntu-only.)

PR #1187, opened 2016-09-24
fs: enable ceph-fuse permission checking for all pjd suites

gregsfortytwo:
As we improve our security model, we need to test it instead of the kernel.


gregsfortytwo:
Whoops, forgot to do this when I merged the UserPerm Ceph branch and probably broke some master runs. :/

PR #1191, opened 2016-09-29

jdurgin:

(cherry picked from commit f28f881bda0ad426bfa9c0e22cc4658b2ce2ec5d)

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -286,6 +286,20 @@ class Thrasher:
             if imp_remote != exp_remote:
                 imp_remote.run(args=cmd)
 
+            # apply low split settings to each pool
+            for pool in self.ceph_manager.list_pools():
+                no_sudo_prefix = prefix[5:]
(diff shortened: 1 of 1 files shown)


jdurgin:
this tests https://github.com/ceph/ceph/pull/11252

smithfarm:
@jdurgin This breaks the infernalis-x upgrade suite: http://tracker.ceph.com/issues/17683

We would need to conditionalize this block of code so it only runs if jewel or higher is installed on the test machine.

jdurgin:
@smithfarm if we still want to run the infernalis-x suite, adding an option to ceph_manager to disable this test and setting that in the infernalis-x suite sounds like a reasonable workaround to me.

This was backported to hammer too, so checking for version is a bit more complicated. Infernalis was only skipped since it's eol.

smithfarm:
@jdurgin Like this? https://github.com/ceph/ceph-qa-suite/pull/1252

PR #1192, opened 2016-09-29

jdurgin:

(cherry picked from commit f28f881bda0ad426bfa9c0e22cc4658b2ce2ec5d)


jdurgin:
this tests https://github.com/ceph/ceph/pull/11253

PR #1193, opened 2016-09-29
suites: update log whitelist for scrub msg

jcsp:
Fixes: http://tracker.ceph.com/issues/16016


jcsp:
Tests https://github.com/ceph/ceph/pull/11136

PR #1195, opened 2016-10-04
rados/basic/tasks/repair_test: expand log whitelist

liewegas:
Depending on scrub timing, we may see anywhere from 1-4 errors.


liewegas:
@dzafman does this look reasonable?  I'm guessing a normally scheduled scrub raced with our test and that's why this run got 2 errors instead of 1 or 4:
http://pulpito.ceph.com/sage-2016-10-03_21:06:11-rados-master---basic-smithi/450914/

dzafman:
@liewegas  I wonder if my fix ceph/ceph-qa-suite#1189 would reduce the resubmitted scrubs.  It is possible that a resubmitted scrub will happen while the next test is being set-up.  This could easily cause an unexpected scrub message in the logs.  

We can do the real fix for this by adding a scrub argument like last_scrub_stamp.  If the last_scrub_stamp has advanced duplicate the scrub request is dropped.

dzafman:
We should put in ceph/ceph-qa-suite#1189 instead of this one.

PR #1197, opened 2016-10-07
jewel: Fixes for rados suite test caused failures and improvements

dzafman:
Backport of  ceph/ceph-qa-suite#1189


dzafman:
Ooops

PR #1199, opened 2016-10-08
do not whitelist "failed to encode"

tchaikov:
newer hammer (0.94.6+) with CRC check will print
 cluster [WRN] failed to encode map e2314 with expected crc"
if CRC does not match. but this should not happen if the peer re-encode
the incrementals without CRC fields if we are upgrading from 0.94.6
to a newer hammer version with the fix of tracker#17386.

Fixes: http://tracker.ceph.com/issues/17386

### DIFF
--- a/suites/upgrade/hammer/newer/0-cluster/start.yaml
+++ b/suites/upgrade/hammer/newer/0-cluster/start.yaml
@@ -4,7 +4,6 @@ overrides:
     - scrub
     - scrub mismatch
     - ScrubResult
-    - failed to encode
     - soft lockup
     - detected stalls on CPUs
     fs: xfs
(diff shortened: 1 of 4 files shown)


tchaikov:
the change on ceph side is posted at https://github.com/ceph/ceph/pull/11372

tchaikov:
tested at http://pulpito.front.sepia.ceph.com:80/kchai-2016-10-08_08:59:22-upgrade-wip-17386-hammer---basic-mira/

tchaikov:
test passed.

liewegas:
included this as part of https://github.com/ceph/ceph-qa-suite/pull/1207

PR #1200, opened 2016-10-10
tasks/rebuild_mondb.py: fill up the caps in the keyring

tchaikov:
we will fail 'ceph-monstore-tool' command if the caps is empty so user
won't assign a key without any caps when rebuilding the monstore.


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-10-10_10:05:13-rados-wip-kefu-testing---basic-mira/

PR #1202, opened 2016-10-11
test_volume_client: check volume size

ajarr:
Check that the total size shown by the df output of a mounted volume
is same as the volume size and the quota set on the volume.


ajarr:
@jcsp looks OK?

jcsp:
Looks :thumbsup:, including in next test run

PR #1204, opened 2016-10-11
suites/upgrade: additional point-to-point Jewel upgrade workaround

dillaman:
Upgrade happens in parallel with cls_rbd test run which can result in
a false-positive test failure.


dillaman:
@yuriw This fixes the race condition between the upgrading and the cls_rbd test running in parallel. Test logs from the run are here: http://qa-proxy.ceph.com/teuthology/jdillaman-2016-10-10_17:12:40-upgrade-master-distro-basic-smithi/466016/teuthology.log

Note: it passes the two cls_rbd test runs but fails later on in LibRadosAio

PR #1205, opened 2016-10-11
upgrade: set kraken osd flag on upgrade

liewegas:


suites/rados/singleton-nomsgr/all/13234.yaml
```diff
@@ -216,6 +216,10 @@ tasks:
     - mon.a
     - mon.b
     - mon.c
+  - exec:
+      mon.a:
+        - ceph osd set require_jewel_osds
```
tchaikov: we can drop this requirement, as require_kraken_osds implies it.

tchaikov:
modulo this nit, lgtm.

PR #1206, opened 2016-10-13
rados: move readwrite test out of thrash matrix

liewegas:
The thrash matrix includes min_size = 2 and size = 2.  This
is fine and good, except that the inevitable slow rados ops
that generates make readwrite.yaml issue timeouts and
fail.  Since the workload isn't meant to tolerate slow
requests, move it into the basic/ group.


liewegas:
@tchaikov @yehudasa ?

tchaikov:
makes sense to me.

PR #1207, opened 2016-10-13
more kraken upgrade fixes

liewegas:


liewegas:
http://pulpito.ceph.com/sage-2016-10-15_14:11:58-upgrade:jewel-x-wip-sage-testing---basic-smithi/

yuriw:
Wow!

PR #1210, opened 2016-10-17
Added jewel-x-singleton

yuriw:


yuriw:
http://pulpito.ceph.com/sage-2016-10-17_16:07:32-rados:upgrade-master---basic-smithi/

liewegas:
http://pulpito.ceph.com/sage-2016-10-17_16:07:32-rados:upgrade-master---basic-smithi/

PR #1217, opened 2016-10-20
mon-seesaw: add mon-seesaw test and mon_seesaw task

tchaikov:
mon_seesaw task replaces a monitor with a newly reployed one, in a
single-mon test including this task OSDs will not be able to connect
to cluster if the tracker#17558 is not fixed on the monitor side.

http://tracker.ceph.com/issues/17558

### DIFF
--- a/suites/rados/singleton/all/mon-seesaw.yaml
+++ b/suites/rados/singleton/all/mon-seesaw.yaml
@@ -1,0 +1,26 @@
+roles:
+- - mon.a
+  - osd.0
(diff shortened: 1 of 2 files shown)


tchaikov:
test passed at http://pulpito.ceph.com/kchai-2016-10-20_03:33:37-rados-wip-kefu-testing---basic-mira/

depends on https://github.com/ceph/ceph/pull/11456

athanatos:
lgtm

PR #1219, opened 2016-10-24
tasks/cephfs: add TestDataScan.test_scan_links

ukernel:


gregsfortytwo:
Do you have a test run of this @ukernel?

ukernel:
only tested locally by vstart runner

PR #1223, opened 2016-10-27
libcephfs1 -> libcephfs2 rename

jtlayton:
Companion patchset to the pull request here:

https://github.com/ceph/ceph/pull/11647

Basically, we're changing the version of libcephfs so we need to update ceph-qa-suite accordingly.


jtlayton:
companion teuthology PR is here:

https://github.com/ceph/teuthology/pull/972

PR #1224, opened 2016-10-27
suites/hadoop: run only on ubuntu

jcsp:
Workaround for http://tracker.ceph.com/issues/16640

Signed-off-by: John Spray john.spray@redhat.com


jcsp:
Should have done this at same time as https://github.com/ceph/ceph-qa-suite/pull/1186

gregsfortytwo:
Not sure why we haven't merged this yet?

PR #1226, opened 2016-10-31
releases/jewel: set require_jewel_osds

dachary:

### DIFF
--- a/suites/rados/singleton-nomsgr/all/11429.yaml
+++ b/suites/rados/singleton-nomsgr/all/11429.yaml
@@ -110,7 +110,15 @@ tasks:
 
   - install.upgrade:
       mon.a: null
-  - ceph.restart: [mon.a, mon.b, mon.c, osd.0, osd.1, osd.2]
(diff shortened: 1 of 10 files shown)


suites/rados/singleton-nomsgr/all/11429.yaml
```diff
@@ -110,7 +110,15 @@ tasks:
...
   - install.upgrade:
       mon.a: null
-  - ceph.restart: [mon.a, mon.b, mon.c, osd.0, osd.1, osd.2]
+  - ceph.restart:
+      daemons: [mon.a, mon.b, mon.c, osd.0, osd.1, osd.2]
+      wait-for-healthy: false
+      wait-for-osds-up: true
+  - exec:
+      osd.0:
+        - sleep 300 # http://tracker.ceph.com/issues/17808
```
yuriw: @jdurgin maybe we should use this timeout in hammer-jewel-x suite as well before setting `ceph osd set require_jewel_osds`?

PR #1227, opened 2016-10-31
ec overwrites tests (and bonus fix to fast reads)

athanatos:


athanatos:
https://github.com/ceph/ceph/pull/11701

PR #1231, opened 2016-11-03
DON'T whitelist crc mismatch

tmuthamizhan:
Removed whitelist for crc mismatch and added changes to cluster config
also the order of upgrades to make sure OSDs are upgraded before MONs

Signed-off-by: Tamil Muthamizhan <tmuthami@redhat.com>


tmuthamizhan:
@liewegas @yuriw : please review

yuriw:
From IRC requirements:  

> (01:13:26 PM) sage: yuriw: Tamil: i heard some talk at openstack that suggested that during a hammer->jewel upgrade, even when doing the osds before the mons, there were osdmap crc mismatches. 
(01:14:07 PM) sage: perhaps we should update the hammer-x upgrade suite (in jewel branch) so that it does osds before mons, and remove the crc mismatch whitelist (we already did this in master) so we can check whether this is the case.. 
(01:14:21 PM) sage: if so it should be a reasonably urgent fix for jewel

yuriw:
@tmuthamizhan I would break this PR in two, one that addresses only @liewegas 's requirements and another with other suites optimizations.

PR #1234, opened 2016-11-06
Fix running cephfs python tests without mgr daemon

jcsp:


jcsp:
(This PR is included on the qa-suite branch in this run: http://pulpito.ceph.com/jspray-2016-11-06_11:15:50-kcephfs:recovery-master-testing-basic-mira/)

jcsp:
It's working here: http://pulpito.ceph.com/jspray-2016-11-06_14:23:27-kcephfs:recovery-master-testing-basic-mira/

(all those failures are from other things, not from the failure during setup that this PR fixes)

gregsfortytwo:
Wow, those recovery tests are not doing well. :/

PR #1241, opened 2016-11-09
Added code for mds invalid fscid test in test_failover.py

ramkiperiy:
Added code for testing invalid fscid test case in cephfs which runs on multimds environment
	it adds invalid fscid to mds and check for cluster daemon status.

Added code for rsync added in tasks

	Adding new code for rsync task. This task can run parallel along with any other tasks.
	it does snap creation and rsync the data to fs mount then removal of snap and data dir, this steps will be done in loop till parallel task execution completes

Signed-off-by: Ramakrishnan <ramakrishnan@dhcp35-178.lab.eng.blr.redhat.com>


jcsp:
add to whitelist

jcsp:
test this please

jcsp:
(FYI those comments I just added are for the jenkins bot to let it build PRs from a new contributor)

tasks/cephfs/test_failover.py
```diff
@@ -570,3 +570,57 @@ def set_standby_for(leader_fs, follower_id):
         fs_b.wait_for_daemons()
         self.assertEqual(set(fs_a.get_active_names()), {mds_a, mds_b})
         self.assertEqual(set(fs_b.get_active_names()), {mds_c, mds_d})
+
+    def test_standby_for_invalid_fscid(self):
```
jcsp: Please separate this new test from the addition of rsync, they should be separate pull requests.

tasks/cephfs/test_failover.py
```diff
@@ -570,3 +570,57 @@ def set_standby_for(leader_fs, follower_id):
...
+            self.mds_cluster.mds_stop(mds_a)
+            self.mds_cluster.mds_fail(mds_a)
+            fs_a.wait_for_daemons()
+        else:
+            self.mds_cluster.mds_stop(mds_b)
+            self.mds_cluster.mds_fail(mds_b)
+            fs_a.wait_for_daemons()
+
+        #Check for active mon status, mon should not be down because of invalid fscid
+        if active_mon_num == len([f.split('.')[1] for f in teuthology.get_mon_names(self.ctx)]):
```
jcsp: the get_mon_names function is not checking if mons are up, it is just telling you about the configured mon roles.  You could use CephManager.get_mon_quorum and check that the number of mons in quorum is equal to the number of mons from teuthology.get_mon_names.

tasks/cephfs/test_failover.py
```diff
@@ -570,3 +570,57 @@ def set_standby_for(leader_fs, follower_id):
...
+            log.info("Invalid fscid Test passed")
+        else:
+            self.assertEqual("Invalid fscid test failed", [])
```
jcsp: Rather than an if statement with an assert in one branch, assert the condition you're actually testing.

For example assuming you have got the list of mons in quorum and the list of mon names in the configuration, you would do something like:
```self.assertEqual(quorum_mons, configured_mons, "Not all mons are in quorum!")```

File: tasks/rsync.py [hunk 4]
```diff
@@ -0,0 +1,146 @@
...
+        self.my_mnt.run_shell(["mkdir", "datadir"])
+        self.my_mnt.run_shell(["mkdir", "datadir/subdir"])
+        self.my_mnt.run_shell(["mkdir", "rsyncdatadir"])
+        self.my_mnt.run_shell(["mkdir", "rsyncdatadir/dir1"])
+
+        while not self.stopping.is_set():
+
+            try:
+                # Create file and add data to the file
+                self.my_mnt.write_test_pattern("datadir/subdir/file_a", self.fileSize * 1024 * 1024)
```
jcsp: Rather than creating a datadir here, take it as a configuration argument so that someone can run this in the background to another task, and point it at the data directory in use by that task.

For example, when someone uses the "workunit" task to run a script, it will create a per-client directory within the filesystem, so someone might use a configuration like this:

```
tasks:                                                                          
- ceph-fuse:
- rsync:
    data_dir: client.0                                                                    
- workunit:                                                                     
    clients:                                                                    
      all:                                                                      
        - kernel_untar_build.sh  
```

tasks/rsync.py
```diff
@@ -0,0 +1,146 @@
...
+                # Rsync snapshot data to rsync directory
+                self.my_mnt.run_shell(["rsync", "-azvh", "datadir/.snap/snap1", "rsyncdatadir/dir1/"])
+
+                # Delete snapshot
+                self.my_mnt.run_shell(["rmdir", "datadir/.snap/snap1"])
+
+                # Delete the file created in data directory
+                self.my_mnt.run_shell(["rm", "-f", "datadir/subdir/file_a"])
+
+            except:
```
jcsp: Never use plain "except" handlers like this, always be specific about which exception class you're handling.

tasks/rsync.py
```diff
@@ -0,0 +1,146 @@
...
+                logging.warning(
+                    "Caught Exception while deleting the snapshot because of stale file, Trying to remove it again.")
+                time.sleep(20)
+                self.my_mnt.run_shell(["rmdir", "datadir/.snap/snap1"])
+                self.my_mnt.run_shell(["rm", "-f", "datadir/subdir/file_a"])
+
+            time.sleep(5)
```
jcsp: Let's make the length of this wait configurable

tasks/rsync.py
```diff
@@ -0,0 +1,146 @@
...
+
+                # Delete snapshot
+                self.my_mnt.run_shell(["rmdir", "datadir/.snap/snap1"])
+
+                # Delete the file created in data directory
+                self.my_mnt.run_shell(["rm", "-f", "datadir/subdir/file_a"])
+
+            except:
+                logging.warning(
+                    "Caught Exception while deleting the snapshot because of stale file, Trying to remove it again.")
```
jcsp: Presumably this is working around an issue that you encountered?  Removing a snapshot is supposed to always work iirc.

tasks/rsync.py
```diff
@@ -0,0 +1,146 @@
...
+
+            try:
+                # Create file and add data to the file
+                self.my_mnt.write_test_pattern("datadir/subdir/file_a", self.fileSize * 1024 * 1024)
+
+                # Create Snapshot
+                self.my_mnt.run_shell(["mkdir", "datadir/.snap/snap1"])
+
+                # Rsync snapshot data to rsync directory
+                self.my_mnt.run_shell(["rsync", "-azvh", "datadir/.snap/snap1", "rsyncdatadir/dir1/"])
```
jcsp: Let's have a boolean configuration flag that selects whether to enable snapshots (current routine) or not.  If snapshots are not enabled, run rsync directly from the datadir, instead of from a snapshot.

tasks/rsync.py
```diff
@@ -0,0 +1,146 @@
...
+
+    if config is None:
+        config = {}
+    assert isinstance(config, dict), \
+        "rsync task accepts dict for running configuration"
+
+    log.info("Create object and start the gevent thread")
+    start_rsync = RSync(ctx, config, logger=log.getChild('rsync'))
+    start_rsync.start()
+    start_rsync_tread = start_rsync
```
jcsp: s/tread/thread/

tasks/rsync.py
```diff
@@ -0,0 +1,146 @@
...
+        self.fs.mon_manager.raw_cluster_cmd("mds", "set", "allow_new_snaps", "true", "--yes-i-really-mean-it")
+
+        # Get mount points objects
+        mounts = [v for k, v in sorted(self.ctx.mounts.items(), lambda a, b: cmp(a[0], b[0]))]
+
+        # Using first client mount point to run rsync test.
+        if len(mounts) >= 1:
+            self.my_mnt = mounts.pop(0)
+        else:
+            logging.exception("No mount point available")
```
jcsp: `logging.exception` is used from exception handlers, it does not raise an exception itself.  In this case I would just do a self.logger.error with the message, and then do an `assert(len(mounts) > 0)`

tasks/rsync.py
ramkiperiy: Yes John, Observed this error "failed to remove ‘datadir/.snap/snap1’: Stale file handle" while testing the code but snap will not be available there in folder, so added this warning message and to make sure snapshot removed from dir added sleep interval & cleared the snap once again which worked. Got this problem 5/20 iterations.

tasks/rsync.py
vasukulkarni: rsync verification is missing, without that it will be difficult to say whether it worked or not, you can simplify this if you use function instead of class and you can use parallel function that uses greenlet in background.

ramkiperiy: Please check this pull request for updated code of rsync https://github.com/ceph/ceph-qa-suite/pull/1247
am facing some problem with this fork unable to pull request in this.

ramkiperiy:
Hi John/Vasu, 
         I am getting some problem while trying to pull updated code in that fork, so created a different pull request.
https://github.com/ceph/ceph-qa-suite/pull/1247

Added code for the comments provided. Please check it

Regards,
Ramakrishnan

ramkiperiy:
Attaching test passed teuthology run log and config yaml for Standby_invalid_fscid test.

[invalid_fscid.yaml.zip](https://github.com/ceph/ceph-qa-suite/files/594355/invalid_fscid.yaml.zip)
[invalid_fscid_teuthology.log.zip](https://github.com/ceph/ceph-qa-suite/files/594354/invalid_fscid_teuthology.log.zip)

jcsp:
This is failing because some other code changed recently in master:
```
2016-11-18T14:15:07.156 INFO:tasks.cephfs_test_runner:======================================================================
2016-11-18T14:15:07.157 INFO:tasks.cephfs_test_runner:ERROR: test_standby_for_invalid_fscid (tasks.cephfs.test_failover.TestMultiFilesystems)
2016-11-18T14:15:07.157 INFO:tasks.cephfs_test_runner:----------------------------------------------------------------------
2016-11-18T14:15:07.157 INFO:tasks.cephfs_test_runner:Traceback (most recent call last):
2016-11-18T14:15:07.157 INFO:tasks.cephfs_test_runner:  File "/home/teuthworker/src/ceph-qa-suite_wip-jcsp-testing-20161118/tasks/cephfs/test_failover.py", line 599, in test_standby_for_invalid_fscid
2016-11-18T14:15:07.157 INFO:tasks.cephfs_test_runner:    fs_a = self.mds_cluster.get_filesystem("cephfs")
2016-11-18T14:15:07.157 INFO:tasks.cephfs_test_runner:AttributeError: 'MDSCluster' object has no attribute 'get_filesystem'
```

Please can you rebase on master (i.e. do "git fetch origin ; git rebase origin/master") and update for the interface change (MDSCluster.get_filesystem is replaced by MDSCluster.newfs)

ramkiperiy:
Hi John, Rebase to master is done, please check now.

jcsp:
I squashed and merged this manually to master (the rebase commits in this PR were kind of confusing)

PR #1247, opened 2016-11-11
Added code for rsync added in tasks

ramkiperiy:
Adding new code for rsync task. This task can run parallel along with any other tasks.
it does snap creation and rsync the data to fs mount then removal of snap and data dir,
this steps will be done in loop till parallel task execution completes
Code can support for workunit task in parallel
Updated code based on review comments
https://github.com/ceph/ceph-qa-suite/pull/1241

Signed-off-by: ramkiperiy <rperiyas@redhat.com>


jcsp:
add to whitelist

jcsp:
ok to test

ramkiperiy:
Attaching passed test logs (teuthology.log & config.yaml).
Scenarios:
      1. normal rsync test without snap
      2. rsync with snap enabled 
      3. rsync with workunit task in parallel.
[normal_rsync.yaml.zip](https://github.com/ceph/ceph-qa-suite/files/594468/normal_rsync.yaml.zip)
[normal_rsync_teuthology.log.zip](https://github.com/ceph/ceph-qa-suite/files/594469/normal_rsync_teuthology.log.zip)
[rsync_workunit.yaml.zip](https://github.com/ceph/ceph-qa-suite/files/594470/rsync_workunit.yaml.zip)
[rsync_workunit_teuthology.log.zip](https://github.com/ceph/ceph-qa-suite/files/594472/rsync_workunit_teuthology.log.zip)
[snap_rsync.yaml.zip](https://github.com/ceph/ceph-qa-suite/files/594471/snap_rsync.yaml.zip)
[snap_rsync_teuthology.log.zip](https://github.com/ceph/ceph-qa-suite/files/594473/snap_rsync_teuthology.log.zip)

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+            assert len(mounts) > 0, self.logger.error('No mount available asserting rsync')
+
+        if self.snapEnable:
+            # Enable snapshots
+            self.fs.mon_manager.raw_cluster_cmd("mds", "set", "allow_new_snaps", "true", "--yes-i-really-mean-it")
+            self.snapDummy = self.dataDir + '/.snap/snap'
+
+        if self.workUnit:
+            assert self.snapEnable == False, \
+                self.logger.error('Workunit task and Snapshot enable cannot co-exist asserting rsync')
```
jcsp: The second argument to assert should just be a string.  The way it's written here, if you hit this assert then the assert's message is going to be whatever the return value of logger.error is (probably None).

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # Create destination directory
+        self.my_mnt.run_shell(["mkdir", "rsyncdir"])
+        self.my_mnt.run_shell(["mkdir", "rsyncdir/dir1"])
+
+        while not self.stopping.is_set():
+
+            if self.snapEnable:
+                # Create file and add data to the file
+                self.my_mnt.write_test_pattern("{}/file_a".format(self.sourceDir), self.fileSize * 1024 * 1024)
```
jcsp: The snapshot mode and setting datadir should not be mutually exclusive -- setting snapEnable should just control whether snapshots are used, not whether a test pattern is written to the source directory.

I'm not sure we're ever going to need the mode where it's calling write_test_pattern at all -- the original purpose of this task is to run behind another workunit, not to write its own files.  If you want to keep this code, then maybe just only run it if no data directory was set.

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+                self.snapShot = self.snapDummy + '{}'.format(self.iteration)
+
+                # Create Snapshot
+                self.my_mnt.run_shell(["mkdir", "{}".format(self.snapShot)])
+                self.iteration += 1
```
jcsp: Iteration does not need to be a class attribute, it could just be local to the function

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+        # Using first mount point of cephfs for run rsync test.
+        if len(mounts):
+            self.my_mnt = mounts.pop(0)
+        else:
+            assert len(mounts) > 0, self.logger.error('No mount available asserting rsync')
+
+        if self.snapEnable:
+            # Enable snapshots
+            self.fs.mon_manager.raw_cluster_cmd("mds", "set", "allow_new_snaps", "true", "--yes-i-really-mean-it")
+            self.snapDummy = self.dataDir + '/.snap/snap'
```
jcsp: "snapdummy" variable is only used one place, so get rid of it and just calculate the path in-place below

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+        try:
+            if self.snapEnable:
+                # Rsync snapshot data to rsync directory
+                self.my_mnt.run_shell(["rsync", "-azvh", "{}".format(self.snapShot), "rsyncdir/dir1/"])
+            else:
+                # Rsync data from data directory to rsync directory
+                self.my_mnt.run_shell(["rsync", "-azvh", "{}".format(self.sourceDir), "rsyncdir/dir1/"])
+
+        except CommandFailedError:
+            self.logger.error('rsync command failed asserting rsync')
```
jcsp: This exception handler seems unnecessary: we're not logging any extra information other than to say where it happened, and that will be indicated in the backtrace when the exception gets logged later.  The _run exception handler will print the backtrace.

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+            # by the time someone does a .get() on this greenlet)
+            self.logger.exception("Exception in do_rsync:")
+            raise
+
+    def log(self, x):
+        """Write data to logger assigned to this MDThrasher"""
+        self.logger.info(x)
+
+    def stop(self):
+        time.sleep(self.runTime)
```
jcsp: Why is this sleep needed?

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+    def _run(self):
+        try:
+            self.do_rsync()
+        except:
+            # Log exceptions here so we get the full backtrace (it's lost
+            # by the time someone does a .get() on this greenlet)
+            self.logger.exception("Exception in do_rsync:")
+            raise
+
+    def log(self, x):
```
jcsp: This function is unused (and unnecessary)

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+            raise
+
+    def do_rsync(self):
+
+        # Get mount objects
+        mounts = [v for k, v in sorted(self.ctx.mounts.items(), lambda a, b: cmp(a[0], b[0]))]
+
+        # Using first mount point of cephfs for run rsync test.
+        if len(mounts):
+            self.my_mnt = mounts.pop(0)
```
jcsp: Let's allow configuring which mount point to use.  Ultimately we will want to run this with two clients, and point the workunit at one and the rsync task at the other.

Currently this is possible by letting the rsync task use client.0 by default, and pointing the workunit at client.1, but it would be better to make it explicit.

jcsp (changes requested):

tasks/rsync.py
ramkiperiy: John, I tried that but mount point and workunit directory will be in testdir which is "/home/ubuntu/cephtest" and snapshot creation can be done in cephfs mount point which will not be having any data because of this used write_pattern.

Snap creation will fail in workunit directory since it is normal linux kernel directory.

Let me know you thoughts, i will change the code if required.

tasks/rsync.py
```diff
@@ -0,0 +1,218 @@
...
+        else:
+            # Create a data directory, sub directory and rsync directory
+            self.my_mnt.run_shell(["mkdir", "{}".format(self.dataDir)])
+            self.my_mnt.run_shell(["mkdir", "{}".format(self.sourceDir)])
+
+        # Create destination directory
+        self.my_mnt.run_shell(["mkdir", "rsyncdir"])
+        self.my_mnt.run_shell(["mkdir", "rsyncdir/dir1"])
+
+        while not self.stopping.is_set():
```
ramkiperiy: Even i faced the problem but not immediately, it was at the end when workunit task completed and removed the directory and rsync was just inside while loop. tried increasing the waitTime at end of while loop not seen the problem.
may be i can change default waittime to 10s

tasks/rsync.py
ramkiperiy: Even i had same idea but if we are not using snapshot scenario then it will be declared without any usage, to avoid it declared it as class attribute if snap enabled.

tasks/rsync.py
ramkiperiy: This will be useful if the user executes rsync as sequential. for eg: if this is not set then rsync while loop iteration happens only once, so added wait time which will make sure rsync runs till runTime.

tasks/rsync.py
ramkiperiy: okay.

tasks/rsync.py
ramkiperiy: i will make this change.

tasks/rsync.py
ramkiperiy: I will update the code only with assert message (second argument)

tasks/rsync.py
jcsp: The important thing is not to wait for a particular length of time, but to wait *until* the directory exists (i.e. keep checking with os.path.exists or similar and proceed when it exists)

tasks/rsync.py
jcsp: > Snap creation will fail in workunit directory since it is normal linux kernel directory.

If you're doing a workunit from client.0 then you'll have /home/ubuntu/cephtest/mnt.0/workunit.client.0, and you can create snapshots in /home/ubuntu/cephtest/mnt.0/workunit.client.0/.snap

tasks/rsync.py
jcsp: Still, it is always better to only declare a variable at the scope that you need it -- if it's only used inside one function, then declare it inside that function

tasks/rsync.py
jcsp: The issue during the end of the task is harder, especially with snapshots: if we have created snapshots then the workunit task will fail when it tries to remove its directory.  We might need to also modify the workunit task to notice if there are snapshots when it tries to remove its directory, and make it wait for those snapshots to go away.

If necessary we could also have the rsync task expose a reference to itself on the global ctx object, and have the workunit task cooperate with it in that way.

For the non-snapshot case, I guess we probably just see the rsync operation itself failing when it tries to run while the workunit directory is removed: in that case we would need an exception handler that checks if the workunit dir is gone, and if it is gone then terminates the rsync task, and if it is still there then reraises the exception.

tasks/rsync.py
ramkiperiy: yes correct without workunit. it will run rsync as normal IO's in particular mount point.

tasks/rsync.py
ramkiperiy: okay, i will make this change.

ramkiperiy:
Attaching test config and log file used for executing rsync data from workunit directory.
[rsync_config.yaml.zip](https://github.com/ceph/ceph-qa-suite/files/617020/rsync_config.yaml.zip)
[rsync_teuthology.log.zip](https://github.com/ceph/ceph-qa-suite/files/617021/rsync_teuthology.log.zip)

ramkiperiy:
Hi John, 
    Any update on this PR ?

liewegas:

PR #1250, opened 2016-11-11
[DNM] rgw: retry on 'radosgw-admin realm pull' failures

cbodley:
the 'realm pull' sometimes fails because the master gateway hasn't
finished starting up - add retries for up to 30 seconds when we see this
error:

CommandFailedError: Command failed on mira032 with status 154: ...


yehudasa:
@cbodley lgtm

cbodley:
doesn't appear to fix the problem.. all 10 tries failed the same way, and the teuthology.log indicated an odd crash. updated http://tracker.ceph.com/issues/17794 with details

cbodley:
radosgw instances are taking over 1min to start in some runs. I raised the time limit from 30s to 5min, and these failures seem to go away

though http://tracker.ceph.com/issues/17872 seems to be another consequence of gateways taking a long time to start up, so there's a deeper issue that needs to be addressed

cbodley:
see https://github.com/ceph/ceph-qa-suite/pull/1269

PR #1251, opened 2016-11-11
distros/supported: add xenial

liewegas:


yuriw:
@liewegas would we then want to remove trusty ?  It will add ~30% load on sepia

liewegas:
i just emailed the list to ask for feedback on that

PR #1256, opened 2016-11-14
upgrade/hammer-x/stress-split: set require_jewel_osds

dachary:
It was missing and the cluster permanently stays on WARNING state after
the upgrade of the OSDs.


dachary:
@yuriw I guess you already run into this one as well ?

dachary:
turns out that does not work as expected http://pulpito.ceph.com/loic-2016-11-14_12:43:45-upgrade:hammer-x-jewel-backports-distro-basic-vps/ failed

PR #1257, opened 2016-11-14
rados/multimon: Rewrite clock skew test

adamemerson:
As we are removing the clock skew parameter, actually skew the system clock.


liewegas:
I think we can easily replace this with an equivalent test.  Combine it with teh other yaml fragments and put it in the rados/singleton/ dir somewhere, put the mon.b on a separate host, and add something like

<pre>
- exec:
    mon.b:
      - date --set ...
</pre>
the hard part is figuring out the current time + 10 seconds...

tchaikov:
@liewegas shall we remove this tests or shall we wait until we have a replacement of this test which does not rely on the clock skew setting?

liewegas:
I would prefer to fix the test now or else we'll forget.  I think it's 
pretty straightforward.. hard part is just writing a shell command to fast 
forward the clock

adamemerson:
I'll make a replacement test, I'm just trying to get my Sepia access fixed, I think I lost my VPN credentials when reinstalling and I've been trying to see if I have them somewhere.

liewegas:

PR #1258, opened 2016-11-15
rados/upgrade: skip libcephfs-devel install

liewegas:


tchaikov:
lgtm. and we can exclude libcephfs2, libcephfs_jni1, libcephfs-java, libcephfs-jni also. as cephfs is not involved in this upgrade test.

PR #1262, opened 2016-11-15
cephfs: get new fsmap after marking clusters down

batrick:
Credit to John Spray for identifying the problem/cause.

Fixes: http://tracker.ceph.com/issues/17894

Signed-off-by: Patrick Donnelly <pdonnell@redhat.com>


batrick:
Test: http://pulpito.ceph.com/pdonnell-2016-11-17_01:42:14-fs:multifs-master-distro-basic-mira/

@jcsp, I think this is ready to be merged.

PR #1264, opened 2016-11-15
Jewel osds before mon

tmuthamizhan:


tmuthamizhan:
tested the change and it looks good.

PR #1266, opened 2016-11-16
Added /upgrade/jewel-x/point-to-point-x

yuriw:


yuriw:
http://pulpito.ceph.com/yuriw-2016-11-16_15:07:00-upgrade:jewel-x:point-to-point-x-jewel-distro-basic-smithi/

PR #1267, opened 2016-11-16
prerequisite to set require_jewel_osds flag is now met

tmuthamizhan:
ALL OSDs in the cluster is now upgraded to jewel prior to
setting require_jewel_osds flag

Signed-off-by: Tamil Muthamizhan <tmuthami@redhat.com>


liewegas:
We only want to upgrade half of the osds at this stage.. teh second half shouldn't get upgraded until later

tmuthamizhan:
ok, so the osds in the first node will be upgraded in firsthalf, osds in the second node followed by mons in first node will be upgraded in secondhalf. i guess that would work.

dachary:
I believe this already is merged as part of another pr

PR #1269, opened 2016-11-16
rgw: fix some races with radosgw and radosgw-agent startup

cbodley:
Startup races are the root cause of several failures in the rgw suite. Teuthology's daemon framework doesn't allow us to block until the daemon finishes startup, so we've added sleeps to account for this. But there are cases where `radosgw` and `radosgw-agent` instances take over a minute to start, so the sleeps of 5-20 seconds are insufficient (and raising their durations would be wasteful in most cases).

This PR replaces these sleeps with http requests that poll the daemon with retry/backoff until it starts accepting connections. For `radosgw`, this happens at the end of `start_rgw()`. For `radosgw-agent`, it just adds retry/backoff to the http requests we're already sending.

Fixes: http://tracker.ceph.com/issues/16129
Fixes: http://tracker.ceph.com/issues/17872

### DIFF
--- a/suites/rgw/singleton/all/radosgw-admin-data-sync.yaml
+++ b/suites/rgw/singleton/all/radosgw-admin-data-sync.yaml
@@ -50,8 +50,6 @@ tasks:
         name: client1-system-user
         access key: 1te6NH5mcdcq0Tc5i8i3
         secret key: Py4IOauQoL18Gp2zM7lC1vLmoawgqcYPbYGcWfXw
-- sleep:
-    duration: 15
 - radosgw-agent:
     client.0:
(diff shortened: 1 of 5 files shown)


oritwas:
great!

oritwas:
we need to backport it to jewel

cbodley:
this ran successfully against a recent branch: http://pulpito.ceph.com/cbodley-2016-11-15_14:35:00-rgw-wip-rgw-compression-type-random---basic-mira/ (the connection failures were caused by http://status.sepia.ceph.com/incident/492)

oritwas:
@yehudasa , can you merge?

PR #1271, opened 2016-11-17
tasks.rgw: 'time' imported but unused

oms4suse:
flake8 was failing.

Signed-off-by: Owen Synge <osynge@suse.com>

### DIFF
--- a/tasks/rgw.py
+++ b/tasks/rgw.py
@@ -6,7 +6,6 @@ import contextlib
 import json
 import logging
 import os
-import time
 import errno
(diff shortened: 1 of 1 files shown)


tasks/rgw.py
```diff
@@ -6,7 +6,6 @@
 import json
 import logging
 import os
-import time
```
tchaikov: the `time.sleep(20)` was removed in 3e68bfdbb29e933edb06f73c88eed176ffacc2e3.

PR #1272, opened 2016-11-17
jewel: rgw: fix some races with radosgw and radosgw-agent startup

cbodley:
(jewel backport of https://github.com/ceph/ceph-qa-suite/pull/1269)

Startup races are the root cause of several failures in the rgw suite. Teuthology's daemon framework doesn't allow us to block until the daemon finishes startup, so we've added sleeps to account for this. But there are cases where `radosgw` and `radosgw-agent` instances take over a minute to start, so the sleeps of 5-20 seconds are insufficient (and raising their durations would be wasteful in most cases).

This PR replaces these sleeps with http requests that poll the daemon with retry/backoff until it starts accepting connections. For `radosgw`, this happens at the end of `start_rgw()`. For `radosgw-agent`, it just adds retry/backoff to the http requests we're already sending.

Fixes: http://tracker.ceph.com/issues/16129
Fixes: http://tracker.ceph.com/issues/17872

### DIFF
--- a/suites/rgw/singleton/all/radosgw-admin-data-sync.yaml
+++ b/suites/rgw/singleton/all/radosgw-admin-data-sync.yaml
@@ -50,8 +50,6 @@ tasks:
         name: client1-system-user
         access key: 1te6NH5mcdcq0Tc5i8i3
         secret key: Py4IOauQoL18Gp2zM7lC1vLmoawgqcYPbYGcWfXw
-- sleep:
-    duration: 15
 - radosgw-agent:
     client.0:
       max-entries: 10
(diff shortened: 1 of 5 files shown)


cbodley:
clean run of jewel branch: http://pulpito.ceph.com/cbodley-2016-11-16_11:34:24-rgw-jewel---basic-mira/

dachary:
@cbodley looks good to me, assuming the flake8 & teuthology run pass.

cbodley:
@dachary i cherry-picked an extra commit to fix the flake error

dachary:
I've rarely seen such a clean run. Excellent :-)

PR #1274, opened 2016-11-17
Improve scrub analysis, list-inconsistent-obj output and osd-scrub-repair test

dzafman:
This needs to go in for rados suite to still work.


dzafman:
This was used to test ceph/ceph#9613 which was recently merged.

PR #1279, opened 2016-11-22
Excluded 'libcephfs2','libcephfs-devel','libcephfs-dev'

yuriw:
Removed firefly and infernalis paths
Removed mds role from hammer (per chat with Jason)
Add "fs: xfs" override


yuriw:
duplicate

PR #1282, opened 2016-11-23
fs: unify common parts of sub-suites

batrick:
To make it easier to work on the multimds suite changes, I split the fs overhaul from: https://github.com/ceph/ceph-qa-suite/pull/1114

Main idea is to unify common parts of sub-suites to make shared configuration obvious and distinguish the differences.


batrick:
I've got an fs suite run in progress here: http://pulpito.ceph.com/pdonnell-2016-11-23_17:30:19-fs-master-distro-basic-mira/

Previous run here: http://pulpito.ceph.com/pdonnell-2016-11-21_19:10:15-fs-master-distro-basic-mira/
4/5 failures fixed by a small correction and retested here: http://pulpito.ceph.com/pdonnell-2016-11-22_19:29:40-fs-master-distro-basic-mira/

Last failure is a bug in libcephfs_interface_test which is reported here: http://tracker.ceph.com/issues/18013

gregsfortytwo (approved):
This looks good to me as I skim through it. I assume we have successful runs of each suite (for the appropriate per-suite meaning of "successful")?

jtlayton:
Potential fix for bug 18013 is here: https://github.com/ceph/ceph/pull/12228

PR #1283, opened 2016-11-24
thrashosds: try ceph-objectstore-tool for one minute

smithfarm:
If ceph-objectstore-tool binary is not present, it's likely because we're in
the middle of an upgrade. Look for the binary for up to one minute instead of
just running it and failing with "exec: ceph-objectstore-tool: not found".

Fixes: http://tracker.ceph.com/issues/18014
Signed-off-by: Nathan Cutler <ncutler@suse.com>

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -18,6 +18,7 @@ from teuthology import misc as teuthology
 from tasks.scrub import Scrubber
 from util.rados import cmd_erasure_code_profile
 from util import get_remote
+from teuthology.contextutil import safe_while
 from teuthology.orchestra.remote import Remote
 from teuthology.orchestra import run
 from teuthology.exceptions import CommandFailedError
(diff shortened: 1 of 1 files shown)


smithfarm:
Pushed `wip-18014` branch to ceph/ceph-qa-suite

smithfarm:
Running 5 thrasher jobs to verify that the PR does not cause an obvious regression:

`./virtualenv/bin/teuthology-suite --priority 101 --suite rados/thrash -l 5 --ceph master --email ncutler@suse.cz --machine-type smithi --suite-branch wip-18014`

**running** http://pulpito.ceph.com:80/smithfarm-2016-11-24_10:38:52-rados:thrash-master---basic-smithi/

smithfarm:
[(teuthology issue#18014) upgrade:hammer-x fails with exception from ceph-objectstore-tool: exp list-pgs failure with status 127](https://trello.com/c/rfean1fj/2208-teuthology-issue-18014-upgrade-hammer-x-fails-with-exception-from-ceph-objectstore-tool-exp-list-pgs-failure-with-status-127)

smithfarm:
@dmick, @dachary Care to review?

dachary:
<pre>
</pre>

I would even make it 10 minutes to rule out even the most improbable race.

smithfarm:
@dachary OK, 10 minutes it is.

liewegas:
We can often do the package upgrade before doing the thrashing or workload (but delay the daemon restarts until when we want them) but that doesn't work for double-upgrades, so this seems okay!

PR #1285, opened 2016-11-26
upgrade/hammer-x: do not whitelist 'failed to encode map'

liewegas:
We should avoid these messages now.


yuriw:
:+1:

PR #1288, opened 2016-11-29
upgrade/hammer-x: verify shec before the full upgrade

dachary:
The hammer-x/stress-split-erasure-code upgrade sequence comes from
hammer-x/stress-split and was modified to fully upgrade the cluster. It
previously upgraded only half of it. Verifying that the shec plugin is
not available and that trying to set it does not crash the OSD or the
MON must be tried before the upgrade is complete.


dachary:
I think that's why http://pulpito.ceph.com/sage-2016-11-26_23:40:14-upgrade:hammer-x-jewel---basic-smithi/579409/ fails.

dachary:
<pre>
teuthology-suite -k distro --verbose --suite upgrade/hammer-x/stress-split-erasure-code --suite-branch wip-shec-upgrade-jewel --ceph jewel --machine-type vps --priority 101 machine_types/vps.yaml
</pre>

* **running**

liewegas:
thanks!

PR #1290, opened 2016-11-29
thrashosds: try ceph-objectstore-tool for 10 minutes

smithfarm:
If ceph-objectstore-tool binary is not present, it's likely because we're in
the middle of an upgrade. Do not try to run the binary until we verify that
it's really present. If it is absent, spend up to 10 minutes waiting for it to
appear.

Before this patch there was quite a large window for a race to occur. This
patch doesn't entirely eliminate it, but drastically reduces it.

Fixes: http://tracker.ceph.com/issues/18014
Signed-off-by: Nathan Cutler <ncutler@suse.com>
(cherry picked from commit 862b47faac1fc9f05ee3322ee4b65cf3d3d666c5)

### DIFF
--- a/tasks/ceph_manager.py
+++ b/tasks/ceph_manager.py
@@ -18,6 +18,7 @@ from teuthology import misc as teuthology
 from tasks.scrub import Scrubber
 from util.rados import cmd_erasure_code_profile
 from util import get_remote
+from teuthology.contextutil import safe_while
 from teuthology.orchestra.remote import Remote
 from teuthology.orchestra import run
 from teuthology.exceptions import CommandFailedError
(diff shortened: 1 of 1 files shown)


smithfarm:
@dachary Sage merged 862b47f (of which this is a cherry-pick) into master recently. Do you see any reason not to merge it into jewel as well?

dachary:
LGTM @smithfarm it would be good to run at least one job using this code path to make sure it actually works. I don't see why not but ... ;-)

smithfarm:
Rebased.

Running 5 thrasher jobs to verify that the patch does not cause an obvious regression:

`./virtualenv/bin/teuthology-suite --priority 101 --suite rados/thrash -l 5 --ceph jewel --email ncutler@suse.cz --machine-type smithi --suite-branch wip-18014-jewel`

**running** http://pulpito.ceph.com:80/smithfarm-2016-12-04_09:09:34-rados:thrash-jewel---basic-smithi/

PR #1291, opened 2016-11-29
rados/thrash*: do no set osd_max_backfills=1

liewegas:
This can lead to a copy-from vs backfill deadlock; see
http://tracker.ceph.com/issues/18085

This effectively reverts 5e880228fd7f59063d22a51eb5488b369b0c8360
which aimed to catch bugs in recovery reservations.


liewegas:
@athanatos is this ok with you?

athanatos:
Hmm, I'd rather have a mix of 3/6/9 rather than leave it at the default.

liewegas:
@athanatos updated... how's this?

PR #1292, opened 2016-11-30
rados: avoid xenial for upgrade tests

liewegas:


djgalloway:
Does this encompass the tests that failed as reported in http://tracker.ceph.com/issues/18089 too?

liewegas:
as far as i can tell

theanalyst:
I think the  symlinks are wrong in the second commit.. pushed https://github.com/ceph/ceph-qa-suite/commit/3e1c302b9df1a689ac70a5b77b33f142027bed69 (as branch jewel-avoid-xenial-upgrade) to run the tests

theanalyst:
I don't think my commit fixed the issue yet, scheduling a suite with that still ended up with the same failure

PR #1293, opened 2016-11-30
qemu: Set permissions on /dev/kvm

djgalloway:
Hack until fix in https://bugzilla.redhat.com/show_bug.cgi?id=1333159 makes it into CentOS repos.

http://pulpito.ceph.com/dgalloway-2016-11-29_00:02:41-rbd-wip-ec-partial-overwrites---basic-smithi/

Signed-off-by: David Galloway <dgallowa@redhat.com>

### DIFF
--- a/tasks/qemu.py
+++ b/tasks/qemu.py
@@ -292,6 +292,13 @@ def run_qemu(ctx, config):
             tdir=testdir,
             client=client
         )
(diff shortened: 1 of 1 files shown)


dmick:
I verified the permissions were correct on all the jobs in the run above, even though some jobs failed (for other reasons)

PR #1296, opened 2016-12-02
upgrade: ceph-test is needed for ceph-coverage 

dachary:


theanalyst:
ack
we have successful runs now

PR #1298, opened 2016-12-02
krbd/unmap: pre-single-major is now 3.12.z

idryomov:
3.13 won't build on newer distros, due to lack of gcc5 support.
nightly/pre-single-major (with a slash) is now gone.

Signed-off-by: Ilya Dryomov <idryomov@gmail.com>

### DIFF
--- a/suites/krbd/unmap/kernels/pre-single-major.yaml
+++ b/suites/krbd/unmap/kernels/pre-single-major.yaml
@@ -1,7 +1,7 @@
 overrides:
   kernel:
     client.0:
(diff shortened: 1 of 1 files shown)


idryomov:
Tested both with both gitbuilders and shaman:
http://pulpito.ceph.com/dis-2016-12-02_15:34:17-krbd:unmap-jewel-testing-basic-smithi/
http://pulpito.ceph.com/dis-2016-12-02_15:56:05-krbd:unmap-jewel-testing-basic-smithi/

PR #1302, opened 2016-12-07
Fixes for PR284

batrick:
I couldn't merge this myself due to master being protected.

https://github.com/ceph/ceph-qa-suite/pull/284

### DIFF
--- a/suites/fs/multiclient/tasks/fsx-mpi.yaml.disabled
+++ b/suites/fs/multiclient/tasks/fsx-mpi.yaml.disabled
@@ -3,6 +3,7 @@ os_type: ubuntu
 os_version: "14.04"
 
 tasks:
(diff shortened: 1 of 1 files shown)


batrick:
@liewegas I don't think John had time to look at this. Do you think it's good to merge?

liewegas:
It looks fine to me!

The real test is to rename it and see if it runs... if so, that woudl be great, as fsx-mpi would be a great thing to have in our qa

batrick:
I did run a test with it enabled but it still failed. I didn't see an obvious way to fix it. Failure message is in this commit message:

https://github.com/ceph/ceph-qa-suite/pull/1302/commits/0a6d76f29b9e6ea6c930655430c183c9d12c4cf1

PR #1303, opened 2016-12-10
tasks/ceph: restore context of osd mount path after mkfs

tchaikov:
in case there are some leftover files with their own SELinux
context, and SELinux will not allow ceph to access these files if
their context is not expected by ceph's SELinux policy.

Fixes: http://tracker.ceph.com/issues/16800


tchaikov:
tested at http://pulpito.ceph.com/kchai-2016-12-09_19:24:35-rados-wip-kefu-testing---basic-mira/

tchaikov:
```
2016-12-10T03:22:22.724 INFO:teuthology.orchestra.run.mira037:Running: 'which restorecon && sudo restorecon -R /var/lib/ceph/osd/ceph-1'
2016-12-10T03:22:22.915 INFO:teuthology.orchestra.run.mira037.stderr:which: no restorecon in (/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/bin)
```
see http://qa-proxy.ceph.com/teuthology/kchai-2016-12-10_02:13:38-rados-wip-kefu-testing---basic-mira/621289/teuthology.log

so the test node does not have restorecon installed. maybe we should not rely on it to reset the context of the files under data directory?

tasks/ceph.py
```diff
@@ -778,6 +778,15 @@ def cluster(ctx, config):
                     '--monmap', monmap_path,
                 ],
             )
+            mnt_point = '/var/lib/ceph/osd/{cluster}-{id}'.format(cluster=cluster_name, id=id_)
+            remote.run(
+                args=[
+                    'which', 'restorecon',
```
tchaikov: restorecon is not in $PATH, just use /sbin/restorecon directly.

PR #1309, opened 2016-12-14
tasks/ceph: restore context of osd mount path before mkfs

tchaikov:
all newly created files and directories under the mount dir inherit the
SELinux type of their parent directory. so we need to set it before
mkfs.

Fixes: http://tracker.ceph.com/issues/16800

### DIFF
--- a/tasks/ceph.py
+++ b/tasks/ceph.py
@@ -752,6 +752,12 @@ def cluster(ctx, config):
                         mnt_point,
                     ]
                 )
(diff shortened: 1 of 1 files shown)


tchaikov:
being tested at http://pulpito.ceph.com/kchai-2016-12-14_11:07:57-rados-master---basic-mira/

athanatos:
Seems fine to me!

PR #1313, opened 2017-10-31
tasks/cephfs: mds_revoke_cap_timeout to mds_session_timeout

jtlayton:
mds_revoke_cap_timeout is being folded into the mds_session_timeout.
Switch to using the mds_session_timeout in the QA tests.

Signed-off-by: Jeff Layton <jlayton@redhat.com>

### DIFF
--- a/tasks/cephfs/test_client_limits.py
+++ b/tasks/cephfs/test_client_limits.py
@@ -120,10 +120,10 @@ class TestClientLimits(CephFSTestCase):
         # Client B tries to stat the file that client A created
         rproc = self.mount_b.write_background("file1")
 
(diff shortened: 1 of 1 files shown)


jtlayton:
This should be merged in conjunction with this PR, to ensure that the tests don't break due to the setting being removed:

https://github.com/ceph/ceph/pull/18274

jcsp:
The ceph-qa-suite stuff moved into qa/ in the ceph.git repository, so I think the copy you want to update is that one.

jtlayton:
Doh! Thanks -- will fix it in the main PR then.

